PR #7, opened 2018-04-17
Improve CircleCI script

mrkkrp:
Key points:

* Keep `ghc` in the repo using git submodules.
* Compile `ghc` with custom config as part of Asterius CI.
* Cache the result and avoid re-compiling GHC unless necessary.
* Pushing to gh pages happens as a separate job using CircleCI workflows and via SSH avoiding compromising GH token.


mrkkrp:
This is ready.

PR #56, opened 2019-01-12
[WIP] macOS + cabal support

angerman:


angerman:
Assuming that the datafilee are in `asterius` instead of `ghc-toolkit`, we can get
`ahc-boot` to pick them up there. 

```
ASTERIUS_DEBUG=True ASTERIUS_DATA_DIR=$PWD/asterius ./bin/ahc-boot
```

Yes. This does conflate the data-dir of asterius and ghc-toolkit, so it's unlikely a perfect solution as is.

Issue #53, opened 2019-01-08
Implement Cabal support


TerrorJack:
## Background

Right now, we assume all user code only imports the boot libs (`ghc-prim`, `base`, etc) and they're provided as plain `.hs` home modules. This doesn't scale in the long run; we need to support compiling and linking `Cabal` packages.

In asterius, there're currently 2 different paths for compiling Haskell code:

* For boot libs, we supply `--with-ghc=ahc` as a `cabal configure` option. `ahc` is a wrapper executable of `ghc` which loads our frontend plugin to intercept the regular pipeline and compile to wasm. Each compilation unit serializes to an object file.
* For input user code, we use GHC API to load them in memory and don't generate any object file.

A reasonable approach for Cabal support: trick `Cabal` to believe it's working with a cross ghc, and we unify the two paths.

## Plan

* [x] Implement `ahc` to be a proper cross-ghc
    * [x] Implement missing wrapper executables: `ahc-pkg`, `ahc-hsc2hs`, `ahc-gcc` and `ahc-ld`.
    * [x] Get rid of the `ASTERIUS_*` environment variables and the "global store" in `ahc`. Instead, the object stores are simply `.a` files stored in the libdir along with package database, and can be reloaded by `ahc` in later runs.
    * [x] Get rid of the old logic for loading Haskell home modules; `ahc-link` is only responsible for generating standalone `.wasm`/`.js` files.
    * [x] `cabal-install` should work at this point; implement unit tests to use `cabal-install` to compile Hackage deps.

TerrorJack:
How `ahc` works & the current conventions of how the wasm object files are emitted and consumed:

1. `Language.Haskell.GHC.Toolkit.FakeGHC` implements the `ghc` wrapping logic: if we're calling it with `--make`, we parse the command line args, modify the `DynFlag`, and finally call into our frontend plugin to do the rest.
2. `Asterius.FrontendPlugin` contains the `ahc` logic: we read the `ASTERIUS_LIB_DIR` environment variable to decide where the wasm object files go.
3. `Asterius.Store` exports the `asteriusModulePath` function which takes the `ASTERIUS_LIB_DIR` value as the root directory and generates the wasm object path, based on the unit id and module name of current module.
4. `ahc` emits wasm objects as well as native objects; the native objects are there to make `Cabal` happy.
[...]

PR #66, opened 2019-02-15
Implement `ahc-ar`

TerrorJack:
Tracking issue: #53 

Previously in #63, we made `ahc` generate `.o` wasm object files which are bundled by gnu ar and copied by Cabal. Now we'll implement `ahc-ar` and other necessary utilities to:

* Add symbol indices into `.a` archives
* Reconstruct stores from archives, without losing lazy loading of individual object files
* Get rid of the legacy object storage completely.

### DIFF
--- a/asterius/app/ahc-ar.hs
+++ b/asterius/app/ahc-ar.hs
@@ -1,17 +1,55 @@
 {-# LANGUAGE ViewPatterns #-}
+{-# LANGUAGE RankNTypes #-}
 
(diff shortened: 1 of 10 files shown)


TerrorJack:
So now our static libs are sufficient for linker's usage. But to get rid of the legacy stores completely we'll need to implement `ahc-ld`, deprecate `runHaskell`/`runCmm` and instead only call `ahc-ld` processes for final linking. Should belong to a separate PR.

PR #71, opened 2019-02-21
Make `ahc-ld` output "executables" which can be quickly converted to node/web artifacts

TerrorJack:
Tracking issue: #53


TerrorJack:
So now we have `ahc-dist` and legacy `ahc-link` sharing most command line arguments and underlying logic. `ahc-dist` takes `--input-exe` and starts from the "executable" generated by `ahc`/`ahc-ld`; `ahc-link` takes `--input-hs` and starts from Haskell sources.

Majority of Cabal support work is done. Yay!

PR #78, opened 2019-02-23
Fix issue with loading wasm file during import

majkrzak:
When module was imported from other directory than this with all generated files, `wasm` file failed to import due to that it was expected to be on the current working dir.


TerrorJack:
Hi! Sorry for the delay; we're occupied on other issues and will review this one before end of week.

PR #95, opened 2019-03-20
Change Cmm function sigs to `[] -> []` and implement tail call opcodes

TerrorJack:
We used to have `[] -> [i64]` sigs for wasm functions translated from a cmm graph. When tinkering on the linker and moving passes back to the codegen, I realized we can change them to `[] -> []`, and the return address can be passed via a global state instead. Benefits compared to the old way:

* We used to reserve a write-once local reg in every cmm function just for the return address. Which was a legacy hack and a waste (see below)
* We used to use the `binaryen` relooper, which easily breaks validation when the graph needs to return anything in an exit block. To workaround that we added a return address reg, make the graph write to that reg and return its content as function output, which (due to misdocumentation) triggered undefined behavior in `binaryen` and led to a long term debugging session. We "solved" that by rolling our own relooper, but this PR proves we can still live with the `binaryen` relooper when that backend is used, thus improving output binary code quality.
* This PR makes it trivial to later add "no-trampolining" mode using tail call opcodes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,6 +61,7 @@ jobs:
             stack --no-terminal test asterius:bigint
             stack --no-terminal test asterius:todomvc
             stack --no-terminal test asterius:cloudflare
+            stack --no-terminal test asterius:fib --test-arguments="--tail-calls"
             stack --no-terminal test asterius:fib --test-arguments="--binaryen --no-gc-sections"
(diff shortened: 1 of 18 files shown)


TerrorJack:
As a consequence, we can now optionally emit wasm binaries with the tail call opcodes, switched on with `--tail-calls`. This gets rid of trampolining during Haskell/Cmm execution.

PR #109, opened 2019-05-09
Add gitignores to all test folders

bollu:
This make git status way more pleasant to use after running the test
suite.


mboes:
I suspect it should be possible to have just one such .gitignore file and achieve the same effect.

bollu:
I wasn't sure about what kind of folder structure we want, and whether the `test/` folder could in the future contain `.js` files. If not, then I can move the `.gitignore` one directory above.

PR #112, opened 2019-05-10
[WIP] Add support for Int64/Int32 -> Int8 coersions

bollu:
Fixes #47

PR #114, opened 2019-05-13
Correct MOD_SS_Conv/Mod_UU_Conv behaviour for I8/I16

bollu:
This PR supersedes #112. This is a cleaned up version of PR #112 with consistent history. 

We add a testcase `bytearraymini` which is a simplified version of `bytearray` which tests `I8/I16 -> I64` upcasting.

Unfortunately, we do not have a testcase for the downcast of `I64/I32 -> I16/I8`, though the code path has been implemented in the same test case.

### DIFF
--- a/asterius/package.yaml
+++ b/asterius/package.yaml
@@ -22,6 +22,7 @@ extra-source-files:
   - test/todomvc/**/*.html
   - test/teletype/**/*.hs
   - test/bytearray/**/*.hs
+  - test/bytearraymini/**/*.hs
   - test/bigint/**/*.hs
   - test/cloudflare/**/*.hs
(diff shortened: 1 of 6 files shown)


TerrorJack:
Good job! I did a little bit of more tweaks before merging:

* Remove the now redundant `__asterius_i32_slot`, since we always use `__asterius_i64_slot` anyway.
* Correct the size of `__asterius_i64_slot`, which should be 8. Currently, we align all data segments by 16 bytes, so earlier code worked since there are redundant padding bytes. But still good to have a correct size here.
* The `wrap*` and `extend*` functions are now inlined, to save the overhead of wasm function calls.
* Misc formatting to improve code style.

PR #116, opened 2019-05-14
[WIP] Change the EDSL API to allow declaring of static sections while building functions

bollu:
We achieve this by having `runEDSL` return an `AsteriusModule` instead of a `Function`. This allows us to 
provide primitives in the `EDSL` such as `declareStaticSymbol`. This would allow us to talk about such a symbol in the context of the function we are building. This helps improve cohesion --- we are declaring and using these static buffers at the same point, instead of declaring them at one place and using them at another.

We introduce `generateWrapperModule`, which wraps all functions in an entire module. This reduces duplication in comparison to the old implementation.
  
The diff is large since most of `Builtins.hs` uses `runEDSL`.

### DIFF
--- a/asterius/src/Asterius/EDSL.hs
+++ b/asterius/src/Asterius/EDSL.hs
@@ -125,6 +125,8 @@ data EDSLState = EDSLState
   , paramBuf :: DList ValueType
   , paramNum, localNum, labelNum :: Int
   , exprBuf :: DList Expression
+  , staticsBuf :: [(AsteriusEntitySymbol, AsteriusStatics)]
+  -- ^ Static variables to be added into the module
   }
 
 initialEDSLState :: EDSLState
@@ -136,6 +138,7 @@ initialEDSLState =
     , localNum = 0
     , labelNum = 0
     , exprBuf = mempty
+    , staticsBuf = mempty
     }
 
 newtype EDSL a =
@@ -171,7 +174,9 @@ runEDSL :: AsteriusEntitySymbol  -- ^ Function name
   -> EDSL ()  -- ^ Builder
   -> AsteriusModule -- ^ Final module
 runEDSL n (EDSL m) =
-  mempty { functionMap=LM.fromList [(n, f)] }
+  mempty { functionMap=LM.fromList [(n, f)]
+         , staticsMap=LM.fromList $ staticsBuf
+         }
   where
     EDSLState {..} = execState m initialEDSLState
     f = adjustLocalRegs $ Function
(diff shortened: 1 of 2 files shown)


bollu:
#### Chat log between me (siddharth) and @TerrorJack  (cheng) describing reasoning of `runEDSL` taking function name as a parameter:
siddharth.bath [11:58]
 @cheng another API decision: `runEDSL <function name> <action>`, or add a primitive into the EDSL called `setFunctionName`? I prefer the former, because it's not really legal to have a function with no name, so it "feels right" to me to be a parameter, but I can see the other argument that it would be more in line with the rest of the DSL to have an action called `setFunctionName`. Thoughts?
Shao Cheng [11:59]
 The latter looks better by me
We reduced one parameter of runEDSL just today, feels a bit odd to add one back
siddharth.bath [12:03]
 Fair.
...
...
siddharth.bath [11:09]
 Also, I have second thoughts about `setFunctionName`. The reason being, it makes `runEDSL` partial, in some sense. We can't create a function without knowing its name, and if someone forgets to call `setFunctionName`, there is _nothing sensible we can do_, right?
I feel it makes the API error prone. This is different from the return values case, where the default is return `void`, and if not, then you set the return value
Shao Cheng [11:10]
 A sensible thing to do is crashing at link-time at least, not deferring this to runtime

siddharth.bath [11:11]
 Right now, the way I've implemented it, the default name is `error`. So that way, when the `asteriusModule` is forced, you'll get an error at least.
But it would still be very annoying to debug, no?
the lack of stack traces...
Shao Cheng [11:12]
 Stack traces can be brought back by making node runner the default
Wait a sec. Don't we already take function name as parameter in your EDSL wip branch? Because we need to return `AsteriusModule` instead @siddharth.bath
siddharth.bath [11:17]
 Yes, we do
No, but I want t change that
but we had agreed that we would have a setter, right?
a setter for the function name, rather than a parameter for `runEDSL`?
Shao Cheng [11:18]
 Yep.
siddharth.bath [11:18]
 Right, so this is the API design that I am having second thoughts about.
simply because, what do we do if the user does not call `setFunctionName`
Shao Cheng [11:19]
 Well, if users can forget to set function name, they can also forget params or return value..function name doesn't look "special"
Dinner time as usual, back here shortly
siddharth.bath [11:21]
 no, my argument is that there is a sensible default for params and return value:
no params, and void respectively
and if the user forgets to set the types correctly and yet say, tries to return a value, binaryen will report an error
unlike setting a function name
there is no sensible default that I can think of
Is there one?
Shao Cheng [11:25]
 And if users forget to set the name, the error is reported even before anything reaches binaryen, and users of EDSL can see an error message with haskell stack trace. So I'd argue that a sensible default isn't really needed in this case
siddharth.bath [11:25]
 by reporting the error using `error`?
Shao Cheng [11:26]
 Yes
siddharth.bath [11:26]
 does that consistently provide a haskell stack trace?
my experience with it has been poor: I am not sure it will actually tell us about the point where `runEDSL` was called
Shao Cheng [11:27]
 We do include `Asterius.Builtins`, which re-exports `IO` with `HasStackTrace => IO`
siddharth.bath [11:27]
 Ah, OK :slightly_smiling_face:
Yes, then it should work.
let me just try this: If I call `error` inside some `runEDSL` block we currently have, we would want the stack trace to have the name of the function where `runEDSL` was invoked, right?
Shao Cheng [11:29]
 The name won't be included, but you shall have the module name and line number, which directs you to the implementation
siddharth.bath [11:30]
 hm, I am not sure how useful that is, because we will be throwing the error from within `EDSL.hs`
Shao Cheng [11:30]
 Oh
Damn I missed this
Let me rethink. If you're pretty sure no extra arguments other than function names will be present in EDSL improvement, I think a single exception makes sense
siddharth.bath [11:32]
 Yes, I think this is the only argument to `runEDSL`, since everything else you can attach to a function has a sensible default, AND will error out nicely if you miss it.
Shao Cheng [11:32]
 I'm just worried that runEDSL is cluttered with a long list of arguments which uglifies the code. Making it look pretty is a strong motivation of implementing EDSL rather than writing raw IR in the first place
siddharth.bath [11:33]
 I understand :slightly_smiling_face:
Shao Cheng [11:33]
 Consensus reached then, great

asterius/src/Asterius/EDSL.hs
```diff
@@ -420,21 +425,25 @@ switchI64 cond make_clauses =
...
+-- |
 -- | Usage:
 -- | runEDSL $ do
--- |   x <- allocStaticBytes "__asterius_i64_slot" (SBS.pack $ replicate 8 0)
--- |   loadI64 x 0
+-- |   x <- allocStaticBytes "x"
+-- |         (Serialized $ SBS.pack $ replicate 8 1)
+-- |   loadi64 x 0
+-- |
+-- |   y <- allocStaticBytes "y" Uninitialized
```
TerrorJack: `Uninitialized 8` in this case.

PR #125, opened 2019-05-16
[WIP] Improve GHC test suite error reporting

bollu:
To be merged into the [`wip-ghc-testsuite`](https://github.com/tweag/asterius/tree/wip-ghc-testsuite).

The output from running `Tasty` using `HSpec` is super noisy, and the prints to `stdout` are completely garbled when using the parallel runner. 

So, we store all the intermediate outputs in an in-memory buffer, which we write to a file at the end of the test run.


bollu:
We can simplify some code if https://github.com/feuerbach/tasty/issues/251 gets a green light, so I can expose some functions from tasty and make our `Ingredient` simpler.

asterius/test/ghc-testsuite.hs
```diff
@@ -37,27 +64,169 @@ getTestCases = do
...
+
+
+-- | Log of tests that have run
+newtype TestLog = TestLog [TestRecord] deriving(Semigroup, Monoid, Generic)
+
+instance ToJSON TestLog where
+
+-- | Append a value to the test log in safe way when we have multiple threads
+consTestLog :: TestRecord -> IORef TestLog -> IO ()
+consTestLog tr tlref = modifyIORef' tlref (\(TestLog tl) -> TestLog $ tr:tl)
```
TerrorJack: `atomicModifyIORef'` please. We might run the tests in parallel.

asterius/test/ghc-testsuite.hs
```diff
@@ -37,27 +64,169 @@ getTestCases = do
...
+runTestCase :: TestCase -> IO ()
 runTestCase TestCase {..} = do
-  _ <- readProcess "ahc-link" ["--input-hs", casePath] ""
+  _ <- readProcess "ahc-link" ["--input-hs", casePath, "--binaryen"] ""
   mod_buf <- LBS.readFile $ casePath -<.> "wasm"
   withJSSession defJSSessionOpts $ \s -> do
     i <- newAsteriusInstance s (casePath -<.> "lib.mjs") mod_buf
     hsInit s i
-    hsMain s i
+    -- | Try to run main. If we throw an exception, return a
```
TerrorJack: It's good to consider the case where the exported `main()` might fail here, and capture it in `RunFailure`. It would be nice to make the error handler here more coarse-grained, to also consider that even `hs_init()`, `newAsteriusInstance`, any logic within the `withJSSession` block might also fail, albeit much less likely than `main()` failures. Moving the `catch` a level up to handle the whole `withJSSession` call is somewhat safer.

And even the logic before the `withJSSession` block is entered might throw. So it might make sense to also add a `CompileFailure e` variant to `RunOutcome`, and use another `catch` to handle it.

asterius/test/ghc-testsuite.hs
```diff
@@ -37,27 +64,169 @@ getTestCases = do
...
+    -- | If the run succeded, now compare outputs.
     hs_stdout <- hsStdOut s i
     hs_stderr <- hsStdErr s i
-    pure (hs_stdout, hs_stderr)
 
+    hs_stdout `shouldBe` caseStdOut
+    hs_stderr `shouldBe` caseStdErr
+
+
+-- | @cheng: Why is this called `makeTestTree`? should it not be called
```
TerrorJack: Because `makeTestTree` produces a `TestTree`, but doesn't really run the test by the time the result is returned in `IO`.

asterius/test/ghc-testsuite.hs
bollu: Right, this makes sense. I'll change the test case accordingly.

asterius/test/ghc-testsuite.hs
bollu: Indeed, realised this later. Will kill the comment, thanks!

asterius/test/ghc-testsuite.hs
bollu: whoops, my bad.

asterius/test/ghc-testsuite.hs
bollu: I think we should rename `runTestCase` to `makeTestCase`? because it too produces what the test case should be...

asterius/test/ghc-testsuite.hs
bollu: Not sure if the way in which I implemented this is optimal. We now have a `CompileOutcome` and `RunOutcome`.

I like this because it feels conceptually cleaner. Thoughts?

TerrorJack: This now looks good to me, thanks for the work.

bollu:
@TerrorJack I'm considering outputting a CSV file instead of (or as well as) a JSON file, because our data is very "row record" style, unlike JSON which is good for heterogeneous data.

asterius/test/ghc-testsuite.hs
```diff
@@ -37,27 +64,189 @@ getTestCases = do
...
+      stats <- computeStatistics smap
+      printStatistics stats time
+      return $ statFailures stats == 0
 
 main :: IO ()
 main = do
-  trees <- getTestCases >>= traverse makeTestTree
-  defaultMain $ testGroup "asterius ghc-testsuite" trees
+  tlref <- newIORef mempty
+  trees <- take 5 <$> getTestCases >>= traverse makeTestTree
```
TerrorJack: We now need to run all cases.

bollu:
Pulled in changes from the other branches here. I'd like to see a green CI here first.

PR #130, opened 2019-05-21
Report unresolved symbols in runtime error messages

TerrorJack:
This PR implements the following mechanism:

* `barf()` as seen in ghc rts. Since we don't have varargs ABI yet, we restrict `barf()` to take exactly one pointer arg, and slice away other args (or provide a `0` in case there's none).
* When the linker sees an unresolved symbol, it dynamically adds an error string as a data segment into the resulting module.
* When the `wasm-toolkit`/`binaryen` backends see an unresolved symbol, they attempt to substitute it with a `barf()` call rather than a simple `unreachable` opcode.

When execution of wasm reaches an unresolved symbol, then the symbol will show up in the js error message.


bollu:
Nice! Could you add a link to `barf` in the GHC RTS along with a note? This should make it easier to understand what's going on.

PR #133, opened 2019-05-22
Add regression test and fix for #60 --- Start building unicode primitives

### LINKED ISSUE #60: Crash using Haskell `read` function
After many tests, I isolated a crash in the wasm execution around the use of the native Haskell function `read`. Same code in standard Haskell works. Here is what I got:

```haskell
module Main where

import Asterius.Types

foreign export javascript "test" test :: IO ()

foreign import javascript "console.log(${1})" consoleLog :: JSString -> IO ()

test :: IO ()
test = do
       let arg = "21";
       let argParsed = read arg :: Int
       let result = argParsed * 2
       let resultSerialized = show result
       consoleLog (toJSString resultSerialized)

main :: IO ()
main = return ()
```
Compilation is fine but execution fails:
```
file:///root/2048/wasm/Main2.mjs:4
  throw err;
  ^

RuntimeError: unreachable
    at wasm-function[1468]:117
    at wasm-function[2318]:171
    at wasm-function[2281]:27
    at wasm-function[2454]:14
    at wasm-function[142]:5
    at file:///root/2048/wasm/Main2.mjs:12:40
```

After many tests, with Strings managed in a several ways. It's the `read` step which fails.

Any idea?

bollu:
- [x] Add a regression test
- [x] Quickly hack up enough unicode primitives treating strings as ASCII to fix this particular regression test
- [ ] Document the full path needed for `read` to work, discuss how best to implement this.\

Fixes #60


bollu:
I realised that there are wrapper helpers: `generateRTSWrapper` that I should call. Will change the code to use this once I get a PR up and running for the float changes.

asterius/src/Asterius/Main.hs
```diff
@@ -463,6 +463,7 @@ ahcDistMain logger task@Task {..} (final_m, err_msgs, report) = do
     withCurrentDirectory (takeDirectory out_wasm) $
     if bundle
       then do
+        putStrLn "BUNDLE: TRUE"
```
TerrorJack: Please remove this line before merging.

asterius/src/Asterius/Builtins.hs
```diff
@@ -1209,6 +1210,24 @@ barfFunction _ =
     s <- param I64
     callImport "__asterius_barf" [convertUInt64ToFloat64 s]
 
+
+unicodeCBits :: [(AsteriusEntitySymbol, (FunctionImport, Function))]
```
TerrorJack: This uses `generateRTSWrapper`, but `generateRTSWrapper` treats `i64` in arguments as *signed*, but results as *unsigned*. I expect all arguments/results here to be *unsigned* though; probably leave a comment regarding sign issue here, or we change `generateRTSWrapper` itself (after carefully reviewing the signedness issue for all usages)

asterius/rts/rts.unicode.mjs
```diff
@@ -0,0 +1,66 @@
...
+    u_iswalpha(code) {
+        return (code >= 'A'.charCodeAt(0) && code <= 'Z'.charCodeAt(0)) ||
+            (code >= 'a'.charCodeAt(0) && code <= 'z'.charCodeAt(0));
+    }
+
+    u_iswalnum(code) {
+        return code >= '0'.charCodeAt(0) && code <= '9'.charCodeAt(0);
+    }
+
+    u_iswupper(code) {
```
TerrorJack: For upper/lower case conversion, we have `String.prototype.toUpperCase()`, please use that instead.

asterius/rts/rts.unicode.mjs
```diff
@@ -0,0 +1,66 @@
+export class Unicode {
+    constructor(logger) {
+        Object.seal(this);
+    }
+
+    u_gencat(code) {
+        if (code >= 'A'.charCodeAt(0) && code <= 'Z'.charCodeAt(0)) {
```
TerrorJack: In general, please use `codePointAt(0)` for converting a single-char js string to `Char#`, and `String.fromCodePoint()` for vice versa. Since the semantics of Haskell `Char#` is unicode code point.

bollu: I'll change the code, but I think this is moot --- none of this code will work if the value held by a `Char#` is not ASCII...

PR #135, opened 2019-05-23
Implement cbits/primFloat.c + some STG helpers for floats in GHC-testsuite

bollu:
Currently experimenting with implementing the very bitfiddly `decodeFloat` function inside the RTS

### DIFF
--- a/asterius/rts/rts.float.mjs
+++ b/asterius/rts/rts.float.mjs
@@ -30,9 +30,9 @@ export class FloatCBits {
         this.FHIGHBIT = 0x00800000;
         this.FMSBIT = 0x80000000;
 
-        // buffer of 4 bytes to hold floats
-        this.bufferFloat = new ArrayBuffer(4);
-        this.viewFloat = new DataView(this.bufferFloat);
+        // buffer of 8 bytes to hold floats/doubles
+        this.buffer = new ArrayBuffer(8);
+        this.view = new DataView(this.buffer);
 
         Object.seal(this);
     }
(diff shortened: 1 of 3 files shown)


asterius/rts/rts.mjs
```diff
@@ -125,6 +128,8 @@ export function newAsteriusInstance(req) {
         emitEvent: ev => __asterius_logger.logEvent(req.events[ev])
       },
       bytestring: modulify(__asterius_bytestring_cbits),
+      // cannot name this float since float is a keyword.
+      float_cbits: modulify(__asterius_float_cbits),
```
TerrorJack: `floatCBits` for consistency with names of other imported modules here, maybe?

asterius/rts/rts.float.mjs
```diff
@@ -0,0 +1,108 @@
...
+        return !isFinite(x);
+    }
+
+    // Does it really make sense to have two functions?  probably not...
+    isDoubleNegativeZero(x) {
+        return Object.is(-0, x);
+    }
+
+    // https://stackoverflow.com/questions/2003493/javascript-float-from-to-bits
+    FloatToIEEE(f) {
```
TerrorJack: How about using `DataView` instead to implement reinterpret-cast from `f32` to `i32`? The code will be cleaner compared to different typed arrays using the same buffer.

Also, since `FloatToIEEE` won't be called in multiple threads, it would be nice to promote the local buffer to a class member to reduce the allocation overhead a bit.

asterius/rts/rts.float.mjs
```diff
@@ -0,0 +1,108 @@
...
+        var buf = new ArrayBuffer(4);
+        (new Float32Array(buf))[0] = f;
+        return (new Uint32Array(buf))[0];
+    }
+
+    // https://stackoverflow.com/questions/2003493/javascript-float-from-to-bits
+    IEEEToFloat(ieee) {
```
TerrorJack: Same comments in `FloatToIEEE`.

asterius/rts/rts.float.mjs
```diff
@@ -0,0 +1,108 @@
...
+        var buf = new ArrayBuffer(4);
+        (new Uint32Array(buf))[0] = ieee;
+        return (new Float32Array(buf))[0];
+    }
+
+
+    __decodeFloat_Int(manp, expp, f) {
+        // https://github.com/ghc/ghc/blob/610ec224a49e092c802a336570fd9613ea15ef3c/rts/StgPrimFloat.c#L215
+        console.log("inside decodeFloat")
```
TerrorJack: Do make sure these debugging logs are removed before a merge.

asterius/rts/rts.float.mjs
```diff
@@ -0,0 +1,108 @@
...
+        console.log("f: ", f);
+        var man, exp;
```
TerrorJack: Please use `let` for local declarations.

PR #141, opened 2019-05-24
Implement support for Debug.Trace.trace

bollu:
- [x] Add a test case
- [x] Implement `debugBelch2`
This PR depends on #133, so the history is understandably messy

### DIFF
--- a/asterius/rts/rts.memory.mjs
+++ b/asterius/rts/rts.memory.mjs
@@ -56,6 +56,19 @@ export class Memory {
   i64LoadU16(p) { return BigInt(this.dataView.getUint16(Memory.unTag(p), true)); }
   heapAlloced(p) { return Memory.unTag(p) >= (this.staticMBlocks << Math.log2(rtsConstants.mblock_size)); }
   strlen(_str) { return this.i8View.subarray(Memory.unTag(_str)).indexOf(0); }
+  strLoad(_str) {
+      let p = Memory.unTag(_str);
+      let s = "";
+      let i = 0;
+      // NOTE: passing a non-null terminated string will give you garbage,
+      // as you deserve.
+      while(1) {
+          let code = this.i8View[p + i];
+          if (code == 0) { return s; }
+          s += String.fromCharCode(code);
+          i += 1;
+      }
+  }
   memchr(_ptr, val, num) {
     const ptr = Memory.unTag(_ptr),
           off = this.i8View.subarray(ptr, ptr + num).indexOf(val);
--- a/asterius/rts/rts.memory.mjs
+++ b/asterius/rts/rts.memory.mjs
@@ -59,15 +59,12 @@ export class Memory {
   strLoad(_str) {
       let p = Memory.unTag(_str);
       let s = "";
-      let i = 0;
-      // NOTE: passing a non-null terminated string will give you garbage,
-      // as you deserve.
-      while(1) {
+
+      for(let i = 0; i < this.strlen(_str); ++i) {
           let code = this.i8View[p + i];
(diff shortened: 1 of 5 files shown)


bollu:
@TerrorJack It compiles! and prints the error :)

asterius/rts/rts.messages.mjs
```diff
@@ -0,0 +1,11 @@
+export class Messages {
+    constructor(memory) {
+        this.memory = memory;
+        Object.seal(this);
+    }
+
+    debugBelch2(fmt, arg) {
```
TerrorJack: iirc all `debugBelch2` calls are `debugBelch2("%s", c_str)`, so it makes sense to simply ignore the first format string here.

asterius/rts/rts.memory.mjs
```diff
@@ -56,6 +56,19 @@ export class Memory {
   i64LoadU16(p) { return BigInt(this.dataView.getUint16(Memory.unTag(p), true)); }
   heapAlloced(p) { return Memory.unTag(p) >= (this.staticMBlocks << Math.log2(rtsConstants.mblock_size)); }
   strlen(_str) { return this.i8View.subarray(Memory.unTag(_str)).indexOf(0); }
+  strLoad(_str) {
+      let p = Memory.unTag(_str);
+      let s = "";
+      let i = 0;
+      // NOTE: passing a non-null terminated string will give you garbage,
+      // as you deserve.
+      while(1) {
```
TerrorJack: Can we use a `for` loop here? We usually use `while(true)` only when the loop range is not trivial to calculate (e.g. walking a stack and stopping for specific frame types)

asterius/rts/rts.messages.mjs
```diff
@@ -0,0 +1,11 @@
...
+        console.error(this.memory.strLoad(fmt), this.memory.strLoad(arg));
```
TerrorJack: `console.error` logs to the `node` process's standard error, keep in mind that we have our own `stderr` and that custom `stderr` buffer is checked by `ghc-testsuite`. So do make sure we log to our own `stderr` as well.

asterius/rts/rts.messages.mjs
bollu: True, but it doesn't _cost_ us to implement a general formatter, and we're correct to this spec, so I don't see a problem per se?

asterius/rts/rts.memory.mjs
bollu: Done.

asterius/rts/rts.messages.mjs
bollu: Done.

asterius/rts/rts.memory.mjs
```diff
@@ -56,6 +56,16 @@ export class Memory {
   i64LoadU16(p) { return BigInt(this.dataView.getUint16(Memory.unTag(p), true)); }
   heapAlloced(p) { return Memory.unTag(p) >= (this.staticMBlocks << Math.log2(rtsConstants.mblock_size)); }
   strlen(_str) { return this.i8View.subarray(Memory.unTag(_str)).indexOf(0); }
+  strLoad(_str) {
+      let p = Memory.unTag(_str);
+      let s = "";
+
+      for(let i = 0; i < this.strlen(_str); ++i) {
```
TerrorJack: This loop traverses the c string twice, first finding its length and then converting to js string. We don't need two traversals.

bollu: Reverting back to old `while(1) {}` style

PR #146, opened 2019-05-27
implement FNV hash instead of MD5

bollu:
We implement a simple hash function to allow `Data.Typeable` to get going, and anything that depends on `GHC.Fingerprint`, really.

We implement a simple `FNV` hash, and hope that this suffices for what we need.

This should fix (at least):
- `test/ghc-testsuite/codeGen/T5129.hs`
- `test/ghc-testsuite/numeric/numrun013.hs`
- `test/ghc-testsuite/deSugar/DsMultiWayIf.hs`

### DIFF
--- a/asterius/rts/rts.memory.mjs
+++ b/asterius/rts/rts.memory.mjs
@@ -42,6 +42,42 @@ export class Memory {
   i32Store(p, v) { this.dataView.setUint32(Memory.unTag(p), Number(v), true); }
   i64Load(p) { return this.dataView.getBigUint64(Memory.unTag(p), true); }
   i64Store(p, v) { this.dataView.setBigUint64(Memory.unTag(p), BigInt(v), true); }
+  i128Load(p) { 
+      console.log("*** load:");
+      // little endian: number with hex digits <0A0B> at address p 
+      // get stored as mem[p] = 0B, mem[p+1] = 0A
+      let low = this.dataView.getBigUint64(Memory.unTag(p), true);
+      console.log("low: ", low);
+      let high = this.dataView.getBigUint64(Memory.unTag(p) + 8, true);
+      console.log("high: ", high);
+      console.log("----");
+      return low | (high << BigInt(64));
+  }
+
+
+  i128Store(p, v) {
+      console.log("*** store v: ", v);
+      // create all 1s of 64 bits.
+      const lowmask = (BigInt(1) << BigInt(64)) - BigInt(1);
+
+      // little endian: number with digits <x y> at address p 
+      // get stored as mem[p] = y, mem[p+1] = x
+      let low = v & lowmask;
+      console.log("v & low: " , low);
+      this.dataView.setBigUint64(Memory.unTag(p), low, true);
+
+      let high = (v >> BigInt(64)) & lowmask;
+      console.log("v & high: " , high);
+
+      let v_recreated = ((high << BigInt(64)) | low);
+      console.log("v_recreated: " , v_recreated);
+      console.assert(v_recreated == BigInt(v));
+
+      // byte addressed, so +8 = 64-bit
+      this.dataView.setBigUint64(Memory.unTag(p) + 8, high, true);
+      console.log("---");
+  }
+
   f32Load(p) { return this.dataView.getFloat32(Memory.unTag(p), true); }
   f32Store(p, v) { this.dataView.setFloat32(Memory.unTag(p), Number(v), true); }
   f64Load(p) { return this.dataView.getFloat64(Memory.unTag(p), true); }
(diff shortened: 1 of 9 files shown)


TerrorJack:
Consider this simple program:

```haskell
{-# LANGUAGE CPP #-}

#include "HsBaseConfig.h"

main :: IO ()
main = print (SIZEOF_STRUCT_MD5CONTEXT :: Int)
```

If run by `ahc-link --input-hs md5.hs --run`, it outputs `0`. So, the configure step of `base` during `ahc-boot` also needs fixing. The `SIZEOF_STRUCT_MD5CONTEXT` constant is used internally in `GHC.Fingerprint`; we must ensure that all such configure-generated constants are correct.

TerrorJack:
In the added unit test, other than struct size, we should actually run the fingerprint functions for one or two buffers and check the result.

bollu:
Generated for a series of whitespaces:

```
a50f5d7c000000000000000000000000
455af67d000000000000000000000000
e5f473eaffffffff0000000000000000
85df7d3b000000000000000000000000
251abc7e010000000000000000000000
c5a4d655010000000000000000000000
657f75f6000000000000000000000000
05aa402a000000000000000000000000
a524e04e000000000000000000000000
45effb55000000000000000000000000
```

TerrorJack:
This PR isn't merge-ready yet;

1. `djb2` hashing, as demonstrated on that webpage, yields a 32-bit hash, yet we require a 128-bit hash here. The chance of collision is much higher.

2. The MD5 context parameter isn't used at all. Consider the case where multiple contexts are initialized, each context is updated multiple times, then finally obtaining the digest for each context, and all the `__hsbase_MD5*` function calls for different contexts might be interleaved. The hash state *must* be properly preserved in the context.

bollu:
Regarding 32-bit, sure. I believe simpler algorithms exist, though.

Regarding saving state, it _is_ saved in the context: 

```
      this.memory.i64Store(ctxp, hash);
```

and:

```
      this.memory.i64Store(digestp, this.memory.i64Load(ctxp));
```

True, this does not care about the C struct's layout, but we are saving our state into the context.

TerrorJack:
You don't need to respect the original C struct's layout, sure, but the usage of context is wrong.

If we call `__hsbase_MD5Update` twice with two chunks, only the second chunk affects the final hash value!

bollu:
Results from the new hash function for consecutive spaces:

```
1df0802e9de806b8777275167c38ad66
ed4950f314da7de45b996504564b429c
bdbe1394f41d6972902e6dd783fc59a7
4d1e02c68d07f778513e3c7dda1c0bb2
dddc201a20f278d29843946f729dbbd7
ad8a08d8ef431898695236c7694c46cf
7d139ddfdac39c0daf10661b51200198
0d47a913a7db42171170efc1cb8e7769
9d998677103e9e216422cadca6d2222a
6d9be1dd7663b5a0d0e7a53232a44933
```

bollu:
We get this failing test:
```
🚨  /tmp/asterius/asterius/test/todomvc/rts.md5.mjs:20:60: Identifier directly after number (20:60)
  18 |   // void MD5Init(struct MD5Context *context);
  19 |   __hsbase_MD5Init(ctxp) {
> 20 |       const offset = 144066263297769815596495629667062367629n;
     |                                                            ^
  21 |       this.memory.i128Store(ctxp, offset);
  22 |   }
  23 | 
```

I can't find a way to enable bigints in babel without installing some plugin. [`babel/plugin-syntax-bigint`](https://babeljs.io/docs/en/babel-plugin-syntax-bigint) seems necessary. Even the [proposals page of babel](https://babeljs.io/docs/en/babel-parser#ecmascript-proposals-https-githubcom-babel-proposals) talks about installing plugins.

I was hoping there was a more lightweight way of doing this `:/`


I'm also unsure how parcel communicates with babel and how this affects babel plugin loading. I'll need to look into this --- I think I'll wait until tomorrow for @TerrorJack since he might have a much clearer picture of this than I do.

asterius/rts/rts.memory.mjs
```diff
@@ -42,6 +42,31 @@ export class Memory {
...
+
+
+  i128Store(p, v) {
+      // create all 1s of 64 bits.
+      const lowmask = (BigInt(1) << BigInt(64)) - BigInt(1);
+
+      // little endian: number with digits <x y> at address p 
+      // get stored as mem[p] = y, mem[p+1] = x
+      const low = v & lowmask;
+      const high = (v >> BigInt(64)) & lowmask;
```
TerrorJack: No need for `& lowmask` here, since the assumption is `v` stays in unsigned 128-bit range, right?

asterius/rts/rts.memory.mjs
```diff
@@ -42,6 +42,31 @@ export class Memory {
...
+
+      console.assert(((high << BigInt(64)) | low) == v);
```
TerrorJack: Come to think of it, if an assertion failure is triggered here, we just print a log and execution carries on, which is not good. I suggest we either remove this line or add a crash with `throw new WebAssembly.RuntimeError("..")`.

PR #152, opened 2019-05-28
Complete implementation of floats

bollu:
The goal of this PR is to fix bugs around float/double handling. So, we should fix:
- [ ] Missing parts of `cbits`
- [x] Incorrect handling of `negate` 

We wish to get the following tests passing from `ghc-testsuite`:
- [x] `codeGen/cgrun044.hs`
- [x] `numeric/numrun014.hs`
- [x] `numeric/numrun012.hs`

### DIFF
--- a/TEST_LOG
+++ b/TEST_LOG
@@ -1,3 +1,17 @@
 Writing this here cause I have no internet at the cafe I am working at...
 
 GHC hangs on: ghc asterius/test/ghc-testsuite/numeric/T15271.hs
+CarryOverflow: "What would a haskell progam be without fibonacci"?
+
+Why can't GHCi handle unboxed stuff?
+
+throwError does not seem to be working. In particular, for example, we had
+an undefined use of MO_Add2 that was not thowing an error. Changing `throwError`
+to `error` fixed this and had us actually raise an error: 
+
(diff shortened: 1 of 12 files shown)


bollu:
This is quite fun. The _printing_ of double is broken, though their hex representation is the same.
##### GHC
```
d: -0.0 | is neg 0: True
f: -0.0 | is neg 0: True
d bytes: 0x8000000000000000
5.0e-324 | hex: 0x0000000000000001 | is denorm: True
2.2250738585072014e-308 | hex: 0x0010000000000000 | is denorm: False
0.0 | hex: 0x0000000000000000 | is denorm: False
```

##### asterius
```
d: -0.0 | is neg 0: True
f: -0.0 | is neg 0: True
d bytes: 0x8000000000000000
0.0e0 | hex: 0x0000000000000001 | is denorm: True
0.0000000000000002e-292 | hex: 0x0010000000000000 | is denorm: True
0.0 | hex: 0x0000000000000000 | is denorm: False
```
Plus clearly, implementation of `denorm` is broken.

The fact that the printing is broken is what bothers me --- this will mean that even after we fix our implementation, the test cases still won't pass, so we need to fix the printing as well.

bollu:
Closing, superceded by splitting this branch and working on the other branches.

PR #156, opened 2019-06-02
Asynchronous JavaScript FFI

TerrorJack:
This PR aims to add support for calling async js functions via the `foreign import javascript` syntax, so users won't need to wrap the Haskell continuation which takes the js return value manually. The usefulness needn't be emphasized, I believe.

Requirements:

* Existing code using JSFFI import/export still works.
* Works on both web/node.
* No `Worker`s or `SharedArrayBuffer`s is involved.


TerrorJack:
Some additional notes:

* The overhead for async calls aren't negligible due to the way we'll implement them (see below). The original plan was assuming all calls are async and using `Promise.resolve` to wrap the js result in generated js script, but now we need to take `unsafe`/`safe`/`interruptible` flags into account.
* The typechecking/desugaring logic of `foreign import javascript` haven't changed. For the emitted raw cmm code, we add a pass which:
    * Scans each cmm basic block and finds "safe" ccalls
    * Splits a basic block body to a sequence of sub-blocks, each sub-block ends with a "safe" ccall
    * Emits context storing/loading code before/after the unconditional branch between sub-blocks
    * Emits an "entry hook" block to override the default entry block, branching to a splitted basic block in case a ccall just returned
* The pass must be run at link-time, since at compile-time there's no easy way to obtain `FFIMarshalState` of all dependent modules.
* The "context" info which we preserve is all of the linear memory (with global regs), plus local regs of that cmm function. All bookkeeping related to async FFI must be done in js, via a set of reserved rts api.
* Other than throwing in js, there's currently no good way of signaling a failure; since this requires allocating a `raise#` closure as the result, which can't be done since FFI import code accepts unlifted results

Alternative approaches which were proved to be hard:

* Using `dsForeignHook` to override the logic of emitting Core bindings for a foreign import. This is the ideal way but sadly I was not enough familiar with this part of ghc (to get it done quickly without all the panics)
* Hacking `CmmLayoutStack` to prevent a `CmmForeginCall` exit node being converted to `CmmUnsafeForeignCall`, so at least we don't need a "block splitter" for safe calls. Again, this proved to be harder than expected.
* "Redirect" the foreign import function closure to a pseudo closure, add special logic in the pseudo closure's entry code to deal with context stuff. But ghc can inline the ccalls, in which case the entry code won't be entered at all.

PR #157, opened 2019-06-03
Implement changes needed for `cgrun044` to pass

bollu:
This fixes double decoding, and implement various bits of the `float` RTS primitives that the test depends on.

@TerrorJack I propose that we teach CI about some ghc testsuite tests which we expect to pass from now on, so that these tests do not regress.

### DIFF
--- a/asterius/rts/rts.float.mjs
+++ b/asterius/rts/rts.float.mjs
@@ -99,11 +99,8 @@ export class FloatCBits {
   isDoubleDenormalized(x) {
     const bits = this.DoubleToIEEE(x);
 
-    console.log("x: %f | bits: %s", x, bits.toString(16));
     const exponent = this.doubleExponentFromBits(bits);
-    console.log("x: %f | exponent: %s", x, exponent.toString(16));
     const mantissa = this.doubleMantissaFromBits(bits);
-    console.log("x: %f | mantissa: %s", x, mantissa.toString(16));
     return exponent === BigInt(0) && mantissa !== BigInt(0);
   }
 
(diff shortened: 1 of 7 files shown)


asterius/rts/rts.float.mjs
```diff
@@ -15,86 +15,229 @@ export class FloatCBits {
...
+    return [low, high];
+  }
+
+  IEEEToFloat(ieee) {
+    this.view.setUint32(0, ieee);
+    return this.view.getFloat32(0);
+  }
+
+  IEEEToDouble(ieee) {
+    this.view.setUint64(0, ieee);
```
TerrorJack: `DataView.prototype.setBigUint64` actually. And don't forget to handle the `i64`/`f64` coercion for `ieee`.

asterius/rts/rts.float.mjs
```diff
@@ -15,86 +15,229 @@ export class FloatCBits {
...
+
+  doubleSignFromBits(bits) {
+    return bits >> BigInt(63);
+  }
+
+  // Check if a double is denormal.
+  isDoubleDenormalized(x) {
+    const bits = this.DoubleToIEEE(x);
+
+    console.log("x: %f | bits: %s", x, bits.toString(16));
```
TerrorJack: Please make sure all `console` calls are removed before merging.

asterius/rts/rts.float.mjs
TerrorJack: Actually it seems `IEEEToDouble` isn't used anywhere. Is it used in subsequent PRs? If not we can pull some weed here.

asterius/src/Asterius/Builtins.hs
```diff
@@ -590,8 +590,13 @@ floatCBits =
        ( AsteriusEntitySymbol func_sym
        , generateRTSWrapper "floatCBits" func_sym param_vts ret_vts))
     [ ("isFloatNegativeZero", [F32], [I64])
+    , ("isDoubleNegativeZero", [F64], [I64])
```
TerrorJack: Are the newly added functions called via `foreign import ccall` somewhere? It seems they are only called via `foreign import javascript` instead, in which case, we don't need to put them into the builtins import; JSFFI handles the relevant import automatically. But if something in `base` has a `ccall` import then we can leave the stuff here.

bollu:
@TerrorJack Made the changes that were requested.

PR #159, opened 2019-06-03
Implement `rintFloat`, rintDouble` fix signedness bug: Allows numrun012 to pass

bollu:
This PR depends on both #158  and #157, so please merge only after these two are green.

### DIFF
--- a/asterius/rts/rts.float.mjs
+++ b/asterius/rts/rts.float.mjs
@@ -138,12 +138,12 @@ export class FloatCBits {
   }
 
   IEEEToFloat(ieee) {
-    this.view.setUint32(0, ieee);
+    this.view.setInt32(0, ieee);
     return this.view.getFloat32(0);
   }
 
   IEEEToDouble(ieee) {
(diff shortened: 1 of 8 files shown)


asterius/rts/rts.float.mjs
```diff
@@ -231,4 +239,178 @@ export class FloatCBits {
...
+    let fexp = BigInt(this.floatExponentFromBits(bits));
+    let fman = BigInt(this.floatMantissaFromBits(bits));
+    let fsign = BigInt(this.floatSignFromBits(bits));
+
+    // put back the float together
+    const reconstructFloat = function() {
+      return this.IEEEToFloat(
+        Number((fsign << BigInt(31)) | (fexp << BigInt(23)) | fman)
+      );
+    }.bind(this);
```
TerrorJack: Using `.bind(this)` is not necessary, use arrow function for `reconstructFloat`/`reconstructDouble`, and their `this` will be fixed to the scope of their definition site.

PR #180, opened 2019-06-12
Fix integerToInt, integerToWord overflows

bollu:
This allows `arith017` to pass. We should in general decide how to deal with these sorts of annoying overflow bugs.


bollu:
This one is quite difficult: When `arith008` fails by default, since at some point, we return `undefined` at `getJSVal` from `decode`. This leads to `undefined / undefined = Nan`, and the errors propogate from there. But the basic error is that we wind up generating `undefined` values.

I don't really understand when this happens or why. But if we switch to strict code that runs inside `IO`, _then the test case passes_. So the bug is _not in the numerics_, it is somewhere else.

I recommend regressing on `arith008`, so that we can get a correct implementation of `integerToWord` and `integerToInt`.

Here is the gist with the pure and impure versions. The impure version succeeds:

https://gist.github.com/bollu/89f0f729faadfcbfaaaf97632518774e

asterius/rts/rts.integer.mjs
```diff
@@ -44,19 +44,30 @@ export class IntegerManager {
...
-
-  integerToWord(i) { return Number(BigInt.asUintN(64, this.decode(i))); }
+      const bi = this.view.getBigUint64(/*offset=*/0, /*little endian=*/true);
+      const n =  this.encode(bi);
+      return n;
+  }
+
+    integerToWord(i, ipiece) { 
+        const n = BigInt.asUintN(64, this.decode(i)); 
+        console.error("n:" , n);
```
TerrorJack: :)

ghc-toolkit/boot-libs/integer-simple/GHC/Integer/Type.hs
```diff
@@ -32,7 +32,7 @@ import GHC.Classes
 import GHC.Magic
 import GHC.Prim
 import GHC.Types
-
+
```
TerrorJack: It would be nice to set up some plugin to trim the trailing whitespaces in your editor, to prevent this in the future.

ghc-toolkit/boot-libs/integer-simple/GHC/Integer/Type.hs
```diff
@@ -69,10 +69,10 @@ wordToInteger w =
     in Integer (js_wordToInteger high low)
 
 integerToWord :: Integer -> Word#
-integerToWord (Integer i) = js_integerToWord i
+integerToWord (Integer i) = or# (uncheckedShiftL# (js_integerToWord i 1#) 32#)  (js_integerToWord i 0#)
 
 integerToInt :: Integer -> Int#
-integerToInt (Integer i) = js_integerToInt i
+integerToInt i = word2Int# (integerToWord i)
```
TerrorJack: Does this work for negative integers as well?

ghc-toolkit/boot-libs/integer-simple/GHC/Integer/Type.hs
```diff
@@ -234,9 +234,9 @@ foreign import javascript "__asterius_jsffi.Integer.smallInteger(${1}, ${2})" js
 
 foreign import javascript "__asterius_jsffi.Integer.wordToInteger(${1}, ${2})" js_wordToInteger :: Word# -> Word# -> Int#
 
-foreign import javascript "__asterius_jsffi.Integer.integerToWord(${1})" js_integerToWord :: Int# -> Word#
+-- | Given integer and which 32-bit _piece_ of the word we want, return that piece. 
+foreign import javascript "__asterius_jsffi.Integer.integerToWord(${1}, ${2})" js_integerToWord :: Int# -> Int# -> Word#
 
-foreign import javascript "__asterius_jsffi.Integer.integerToInt(${1})" js_integerToInt :: Int# -> Int#
```
TerrorJack: Can we do the same for `integerToWord`/`integerToInt` and not rely on `int2Word#` or `word2Int#`? Won't be too much work and I sense some danger in the coercing.

PR #182, opened 2019-06-14
Use runhaskell to generate .stderr and .stdout files

timjb:


asterius/test/ghc-testsuite.hs
```diff
@@ -35,11 +33,6 @@ import Data.Word
 
 -- Much of the code is shamelessly stolen from:
 -- http://hackage.haskell.org/package/tasty-1.2.2/docs/src/Test.Tasty.Ingredients.ConsoleReporter.html#consoleTestReporter
---
```
TerrorJack: Can you split the modifications to `stack.yaml` and the original `ghc-testsuite` test to a separate PR? To keep this one simple and close to the PR topic.

PR #191, opened 2019-06-18
[WIP] numeric tip-of-tree to track number of failures

bollu:
**DO NOT MERGE**

This is a branch to contain the current updated state of numerics so we can monitor how many tests fail currently


bollu:
Closed since most of numerics is now stable.

PR #196, opened 2019-06-21
Implement StableName related primitives

bollu:


asterius/rts/rts.threadpaused.mjs
```diff
@@ -5,7 +5,7 @@ export class ThreadPaused {
   constructor(memory, info_tables, symbol_table) {
     this.memory = memory;
     this.infoTables = info_tables;
-    this.symbolTable = symbol_table;
+
```
TerrorJack: Why is this.symbolTable removed?

bollu:
Ping @TerrorJack

asterius/rts/rts.gc.mjs
```diff
@@ -500,6 +501,23 @@ export class GC {
...
+
+    // Stage the movement of stable pointers. 
+    // Step 1: Move all the pointers
+    // Step 2: Update the pointer -> stablepointer mapping
+    // We cannot do this at the same time, since moving the pointer while
+    // we walk the ptr2stable map can yield an infinite loop:
+    // eg. (ptr:0 stablename: 42) --MOVE--> (ptr:1 stablename:42) --MOVE--> (ptr:2 stablename:42) ...
+    let ptr2stableMoved = new Map();
+    for (const[ptr, stable] of this.stableNameManager.ptr2stable.entries()) {
+      const ptrMoved = this.evacuateClosure(ptr);
```
TerrorJack: We should also evacuate the StgStableName object before adding an entry in the updated mapping.

asterius/rts/rts.gc.mjs
```diff
@@ -500,6 +501,23 @@ export class GC {
...
+      ptr2stableMoved.set(ptrMoved, stable);
+    }
+    this.stableNameManager.ptr2stable.clear();
```
TerrorJack: We can use Object.seal instead of Object.freeze in stable name manager class, then directly assign the new mapping instead of clearing and adding updated entries.

asterius/rts/rts.threadpaused.mjs
bollu: Not used to emacs yet, trying to learn it ;)

PR #197, opened 2019-06-24
Teach GHC testsuite about whitelists and blacklists

bollu:
This has two uses:
1. It allows us to create test trees that run a sub-part of the test, so that we can make use of the parallelism in `tasty`, without having to invoke the tests one-by-one
2. It allows us to run the same check that runs on the CI locally as well.


bollu:
This is currently WIP.

bollu:
@TerrorJack Ping, review requested.

asterius/test/ghc-testsuite.hs
```diff
@@ -224,34 +234,94 @@ consoleOutput tlref toutput smap =
...
-      printStatistics stats time
-      return $ statFailures stats == 0
+serializeToDisk tlref =
+  TestManager [Test.Tasty.Options.Option (Proxy :: Proxy PatternFilePath) ] $
+      \opts tree -> Just $ do
+          let (PatternFilePath patf) = lookupOption opts
+          treeFiltered <- case patf of
+                      Nothing -> return $ tree
+                      Just fpath -> do
+                           cwd <- getCurrentDirectory -- use full path so we get better errors.
```
TerrorJack: To use absolute path in subsequent calls, better to use `System.Directory.makeAbsolute`, rather than getting current directory then concating, so we can support relative/absolute paths out of the box.

asterius/test/ghc-testsuite.hs
```diff
@@ -224,34 +234,94 @@ consoleOutput tlref toutput smap =
...
+        else TestGroup (name <> "-DISABLED") []
+    go (TestGroup name tt) =
+      TestGroup (name <> "-FILTERED") (map go tt)
+    go _ = error $ "unknown test tree type"
+
+-- | Option that describes whether a test file with
+-- white & black lists has been provided. Modelled after
+-- TestPattern
+newtype PatternFilePath = PatternFilePath (Maybe FilePath)
+  deriving(Typeable, Show, Eq)
```
TerrorJack: `Typeable` deriving is on by default since a long time ago, no need to make it explicit here.

PR #201, opened 2019-07-01
[WIP] Fixing T9001 GC crash

bollu:


bollu:
Current status:

The crash occurs at a configuration where we have allocated the blocks:
```js
all_bds:  [
  9007160601084160,
  9007160602132736,
  9007160603181312,
  9007160604229888
]
```

and we detect that the blocks with live objects are:
```js
sorted_bds:  [ 9007160602132736, 9007160604229888 ]
```

and we therefore free the segment:

```js
->>>> freeSegment: i:  0 bd:  9007160602132736  |bd[i+1]:  9007160604229888 
	ix in all:  1 
	l_start:  9007160602148864  l_blocks:  252 l_end;  9007160603181056 
	r:  9007160603181056  r_prev:  9007160604229632 r_prev - r:  1048576
```

This leads us to later crash at:

```
[ERROR] RuntimeError: XXXX Invalid address 0x424242424242425a raw: 4774451407313060442|i32| tag:1111638594
```

The pattern is `0x424242...` because I changed the `memset` to write in `42` everywhere, so I can be sure that this memory that was _wiped after the block allocator_. 

That's the current status. I'm trying to figure out what live object it is that we are accidentally wiping away.

bollu:
we see closure types `9`, `38`, and `32` on the heap, and `4`, `3`, `21` on the stack. So I'm going to spend time debugging the GC walking on those closure types.

Enum stolen from [ClosureTypes.h(https://gitlab.haskell.org/ghc/ghc/blob/master/includes/rts/storage/ClosureTypes.h)

```
#define FUN_1_0                       9
#define RET_FUN                       32
#define BLACKHOLE                     38
```

```
#define CONSTR_2_0                    4
#define CONSTR_0_1                    3
#define THUNK_STATIC                  21
```

bollu:
Bisect on the number of objects we need to have a crash. Turns out that when we go from `505` steps to `506`, we crash. This is in reference to `T9001.hs`. I'm going to now dump out the full heap for `505` and `506` and view the diff

bollu:
Without protecting the first megablock, here is what the crashes look like:

```
-- 300 +
-- 350 +
-- 355 +
-- 356 +
-- FAILURE HERE
-- 357 -
-- 360 -
-- 375 -
-- 400 -
-- 500 -
```

bollu:
This is wild, we seem to have `NaN` for the memory locations.

bollu:
So the types of pointers that we have in the first `64` bytes of the crashing megablock are `38 (BLACKHOLE)` and `9(FUN_1_0)`.

The reason I suspect the first couple objects is that for whatever reason, if I only free the first `32` bytes of the offending megablock, we succeed the test. But if I free the `33`rd byte, then we're screwed. So, either:

- we reach this object and we do not GC it correctly
- We don't even reach this object.

possibiity 2 is worse than possibility 1, so I'd like to rule out possibility 1 first,

PR #206, opened 2019-07-03
[WIP] fix GC crash in CmpWord8

bollu:


bollu:
Note that simply disabling the free list is not enough to make this test pass. It times out after 30 seconds without the freelist allocation. Now running it _Without a timeout_ to see how long it will run.

bollu:
It takes `131s`

```
asterius ghc-testsuite
  test/ghc-testsuite/primops/CmpWord8.hs
    test/ghc-testsuite/primops/CmpWord8.hs: OK (131.01s)
```

PR #207, opened 2019-07-03
Add error messages to the GC runtime

bollu:
We had `throw new WebAssembly.RuntimeError()` which is less valuable than having an error message. So add error messages to the code.


asterius/rts/rts.gc.mjs
```diff
@@ -175,7 +175,7 @@ export class GC {
               break;
             }
             default:
-              throw new WebAssembly.RuntimeError();
+              throw new WebAssembly.RuntimeError("unhandled closure type: " + type);
```
TerrorJack: I recommend using JS template literals as demonstrated in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals. It looks nicer than string concats.

PR #218, opened 2019-07-10
Implement YOLO mode for the runtime

TerrorJack:
This PR implements a `--yolo` switch for `ahc-link` which enables a YOLO mode for the runtime:

* The garbage collector entry function `gcRootTSO` becomes a no-op.
* Mblocks, StablePtr/StableName, JSVal...are never cleaned up.
* Upon a `HeapOverflow` situation, the runtime simply moves to a new nursery.

This will surely exhaust memory in the long run, but it's a useful way to quickly determine whether a runtime bug is rooted in the storage manager. E.g. `T8726` passes if YOLO mode is on.


TerrorJack:
By diffing CI logs of `ghc-testsuite` runs without/with yolo mode, the following failures are "fixed" by switching on `--yolo`, thus indicating the bug is in gc:

* `simplrun004`
* `CmpWord8`
* `T15226`
* `arith008`
* `T8726`
* `cgrun026`
* `T9001`
* `arr020`
* `arr017`

PR #219, opened 2019-07-10
Switch alllocator to use a "freelist of segments"

bollu:
We change the freelist in the mblockallocator to become a list of "free segments". This eliminates error-prone bounds checking, and should be a more stable base for the megablock allocator. This should fix `CmpWord8`

### DIFF
--- a/asterius/rts/rts.mblockalloc.mjs
+++ b/asterius/rts/rts.mblockalloc.mjs
@@ -56,6 +56,7 @@ export class MBlockAlloc {
       const blocks = Math.floor((r - start) / rtsConstants.block_size);
     
       if (req_blocks < blocks) {
+        console.log("allocating from free segment");
         this.memory.i32Store(bd + rtsConstants.offset_bdescr_blocks, req_blocks);
         const rest_bd = bd + (rtsConstants.mblock_size * n);
         assertBdescrAligned(rest_bd);
@@ -87,10 +88,12 @@ export class MBlockAlloc {
       }
 
       if (req_blocks == blocks) {
+        console.log("allocating from free segment");
         this.freeSegments.splice(i, 1);
         return bd;
       }
     }
+    console.log("allocating from new segment");
     const mblock = this.getMBlocks(n),
           bd = mblock + rtsConstants.offset_first_bdescr,
           block_addr = mblock + rtsConstants.offset_first_block;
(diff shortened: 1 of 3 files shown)


asterius/rts/rts.mblockalloc.mjs
```diff
@@ -1,13 +1,25 @@
 import * as rtsConstants from "./rts.constants.mjs";
 import { Memory } from "./rts.memory.mjs";
+import * as assert from "assert";
```
TerrorJack: This is a node-only import which will break the browser target again. We can have a local no-op assert function here for now, before we figure out how to properly do rts assertions.

asterius/rts/rts.mblockalloc.mjs
```diff
@@ -65,25 +85,26 @@ export class MBlockAlloc {
...
-      const bd = l_end + rtsConstants.offset_first_bdescr,
-            start = l_end + rtsConstants.offset_first_block,
-            blocks = (r - start) / rtsConstants.block_size;
-      this.memory.i64Store(bd + rtsConstants.offset_bdescr_start, start);
-      this.memory.i64Store(bd + rtsConstants.offset_bdescr_free, start);
-      this.memory.i64Store(bd + rtsConstants.offset_bdescr_link, 0);
-      this.memory.i32Store(bd + rtsConstants.offset_bdescr_blocks, blocks);
-      this.freeList.push(bd);
+        // memset a custom number purely for debugging help.
+        this.memory.memset(l_end, 0x42 + i, r - l_end);
```
TerrorJack: Better to set to a larger byte e.g. 0xff so it's more likely to get an overflowing address.

asterius/rts/rts.mblockalloc.mjs
```diff
@@ -30,31 +42,39 @@ export class MBlockAlloc {
...
-    const req_blocks = ((rtsConstants.mblock_size * n) - rtsConstants.offset_first_block) / rtsConstants.block_size;
-    for (let i = 0; i < this.freeList.length; ++i) {
-      const bd = this.freeList[i],
-            blocks = this.memory.i32Load(bd + rtsConstants.offset_bdescr_blocks);
+    const req_blocks = 
+          Math.ceil(((rtsConstants.mblock_size * n) - rtsConstants.offset_first_block) / 
+          rtsConstants.block_size);
+
+    for (let i = 0; i < this.freeSegments.length; ++i) {
+      const [bd, r] = this.freeSegments[i];
```
TerrorJack: So, we are saving the (bd, right) tuple as a segment here, which in fact is not too far from the original design which only stores bd. Further more, the freeSegment function below still initializes the bdescr of freed space.

For simplicity, I think we should:

* Store (left, right) tuple, ensure left and right satisfy mblock alignment.
* freeSegment only takes care of wiping bytes.
* When allocMegaGroup picks up a tuple, it initializes the bdescr fields before returning.

asterius/rts/rts.mblockalloc.mjs
bollu: If we memset `0xff + i`, then we can't eyeball the bit-pattern to figure out which segment it came from, unfortunately.

PR #222, opened 2019-07-12
Turn the scheduler into a background daemon

TerrorJack:
This PR is the first step towards multi-threading in asterius: turn the scheduler into a background daemon.

The `rts_eval*` functions are the exposed rts interfaces for evaluation. Previously, each call to `rts_eval*` maps to a call to the scheduler loop which returns asynchronously. However, it's apparent that:

* The scheduler loop is non-reentrant, and a `rts_eval*` call cannot be made before the previous one returns
* During evaluation, we may encounter other evaluation requests (e.g. from `fork#`)
* When a `rts_eval*` call is supposed to return, there may be ongoing background evaluation work.

To better coordinate the evaluation of multiple closures, we now daemonize the scheduler loop:

* A `Channel` JavaScript class is implemented for passing evaluation requests.
* The `Exports` class constructor kicks off a scheduler loop immediately. The loop `await`s requests from the `Channel`.
* All `rts_eval*` calls now simply send requests to the `Channel`, and the result will be available asynchronously.

Since we now work with a single global scheduler loop, it'll be much easier to do the bookkeeping for multiple threads.

### DIFF
--- a/asterius/rts/rts.channel.mjs
+++ b/asterius/rts/rts.channel.mjs
@@ -1,0 +1,17 @@
+export class Channel {
+  constructor() {
+    this.init();
+    Object.seal(this);
+  }
+
+  init() {
+    this.promise = new Promise((resolve, reject) => {
+      this.resolve = resolve;
+      this.reject = reject;
+    }).then(r => (this.init(), r), err => (this.init(), Promise.reject(err)));
+  }
+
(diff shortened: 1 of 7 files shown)


hsyl20:
This PR can be closed now. The relevant bits have been merged.

Issue #257, opened 2019-09-10
Include the missing boot libs


TerrorJack:
As a prequel of #256 

Currently, our boot libs are still missing some components (e.g. `text`). The missing boot libs should be added, with their own unit tests when possible, one package per pull request.

The list of known missing boot libs:

* [ ] `Cabal-2.4.0.1`
* [ ] `ghc-8.6.5`
* [ ] `ghc-compact-0.1.0.0`
* [ ] `haskeline-0.7.4.3`
* [ ] `hpc-0.6.0.3`
* [ ] `libiserv-8.6.3`
* [x] `parsec-3.1.13.0`
* [ ] `process-1.6.5.0`
* [ ] `stm-2.5.0.0`
* [ ] `terminfo-0.4.1.2`
* [x] `text-1.2.3.1`

TerrorJack:
Other than `ghc`, everything else is either installed by `ahc-boot` or `ahc-cabal`, and we don't need to checkin all the boot libs in our tree. So closing this one.

PR #265, opened 2019-09-10
Add "text" boot-lib (#257)

hsyl20:
Add support for "text" library


asterius/rts/rts.unicode.mjs
```diff
@@ -29,7 +29,7 @@ export class Unicode {
     }
 
     u_towtitle(code){
-        return u_toupper(code);
+        return this.u_towupper(code); //FIXME
```
TerrorJack: Would be good to open a related ticket for this one, or all future FIXME or TODO comments.

hsyl20: I have opened https://github.com/tweag/asterius/issues/274

PR #276, opened 2019-09-11
Add `parsec` boot package (#257)

hsyl20:


hsyl20:
> I don't see the real `parsec.hs` containing the test?

Oops! It wasn't added...

PR #282, opened 2019-09-12
Add documentation

hsyl20:
Add some comments and documentation about memory management.

### DIFF
--- a/docs/exec_model.md
+++ b/docs/exec_model.md
@@ -1,0 +1,159 @@
+# Execution model
+
+## JavaScript execution model
+
+JS engines maintain a set of tasks. Among them, the runnable tasks are in a
+queue (macro-task queue). Each task of the queue is executed until its
+completion without any kind of preemption. A long running task may "block" the
+execution of other tasks. This is especially noticeable in a Web browser when
+the DOM rendering task and the UI event handler tasks are blocked.
+
+![JS execution model](js_exec_model.svg)
+
+Each task is in fact composed of a queue of "micro-tasks". Initially there is a
+single micro-task in the queue. This task can queue other micro-tasks with
+``queueMicrotask(task)``. These micro-tasks are executed in FIFO order. When the
+micro-task queue is empty, the macro-task has completed.
+
+**Promises**
+
+[Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise)
+are JS synchronization objects. When they settle they can either be fulfilled or
+rejected.
+
+```
+p = new Promise((resolve,reject) => whatever)
+```
+
+``whatever`` is executed synchronously. ``resolve`` and ``reject`` are functions to set the Promise value.
+
+```
+p.then(r => doSomething)
+p.catch(r => doSomething)
+```
+
+``doSomething`` is always executed asynchronously.
+
+In a micro-task we detect if ``p`` is settled: if it is, we execute ``doSomething``,
+otherwise ``doSomething`` will be executed in a future macro-task (when ``p`` gets
+settled)
+
+```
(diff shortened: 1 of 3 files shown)


hsyl20:
Incidentally this blog post just released is related: https://www.tweag.io/posts/2019-09-12-webassembly-internship.html

docs/exec_model.md
```diff
@@ -0,0 +1,154 @@
...
+b = a.then(g)
+```
+so we queue several micro-tasks in order just like before. If we only have
+settled Promises, the execution is very fast as no macro-task creation/switching
+occurs.
+
+``Promise.all(xs)`` waits for all the promises to settle.
+``Promise.race(xs)`` returns the first settled Promise in xs (eiter resolved or rejected)
+
+With ES2016 we can somewhat replace Promises with await/async:
```
TerrorJack: ES2017 actually, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#Specifications

docs/exec_model.md
```diff
@@ -0,0 +1,154 @@
...
+
+JS supports some kind of multi-threading via
+[Workers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API). A
+Worker has its own task queues and execution engine. Each worker has its own JS
+context and Workers are executed concurrently (e.g. one OS thread per Worker).
+
+Communication between Workers is done via message passing. Using a shared memory
+was possible through
+[SharedBufferArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer)
+but it has been disabled for security reasons.
```
TerrorJack: `SharedArrayBuffer` always works on node, and also in latest versions of Chrome.

docs/exec_model.md
```diff
@@ -0,0 +1,154 @@
...
+[proposal](https://github.com/WebAssembly/threads/blob/master/proposals/threads/Overview.md)
+to add support for threads in WebAssembly reusing Worker and SharedBufferArray
+but it isn't implemented yet.
+
+## Haskell execution model
+
+Haskell threads are represented by a Thread State Object (TSO). Each TSO has its
+own Stack.
+
+In the native RTS, TSOs are executed by Capabilities (virtual CPUs). Each
```
TerrorJack: It's worth noting that the `Task` and `InCall` concepts do not exist in our own runtime; we have a single global `MainCapability` just like the non-threaded runtime. Also, we don't implement the "run queue" yet, and even if we do, it's likely we'll do it in JavaScript instead of using the TSO queue objects on the wasm heap.

docs/exec_model.md
```diff
@@ -0,0 +1,154 @@
...
+### Step 1: cooperative multi-threading
+
+We could introduce a run queue of TSOs. Each time a TSO calls back into the
+scheduler we can schedule another TSO.
+
+We can postpone the execution of the next TSO with
+``setTimeout(executeNextTSO,0)`` in order to let the JS engine execute other
+macro-tasks in between. E.g. JSFFI using ``setTimeout``, browser events, etc.
+
+### Step 2: cooperative multi-threading + Worker
```
TerrorJack: Even when the Spectre mitigation is complete in all major browsers, we should be careful about `Worker` usage and not rely on it by default; it's still heavyweight imho. Running everything in the main thread is not going to block execution of other JavaScript functions noticeably, because our runtime is async by nature and has a lot of chances of yielding control to other functions (e.g. ongoing async calls)

hsyl20:
> Nice writeup about JS/HS execution model. Would be nice to also have reference links to MDN and GHC commentary, there are also some well-crafted materials about execution models on those places

Thanks for the review! I'll try to add some links.

PR #297, opened 2019-09-16
Template Haskell support

TerrorJack:


asterius/src/Asterius/FrontendPlugin.hs
```diff
@@ -31,14 +32,25 @@ frontendPlugin = makeFrontendPlugin $ do
...
+                    GHC.readIServHook = Just asteriusReadIServ,
+                    GHC.writeIServHook = Just asteriusWriteIServ,
+                    GHC.stopIServHook = Just asteriusStopIServ
+                    }
               }
+  do
+    dflags <- GHC.getSessionDynFlags
+    void
+      $ GHC.setSessionDynFlags
+      $ dflags {GHC.settings = (GHC.settings dflags) {GHC.sPgm_i = "false"}}
```
hsyl20: what is the `sPgm_i` option and why do we need to set it?

asterius/src/Asterius/GHCi/Internals.hs
```diff
@@ -0,0 +1,415 @@
...
+
+asteriusReadIServ ::
+  forall a. (Binary a, Typeable a) => GHC.HscEnv -> GHC.IServ -> IO a
+asteriusReadIServ _ _ = withMVar globalGHCiState $
+  \s -> let Just (_, p, _) = ghciJSSession s in readPipe p get
+
+asteriusWriteIServ ::
+  (Binary a, Typeable a) => GHC.HscEnv -> GHC.IServ -> a -> IO ()
+asteriusWriteIServ hsc_env i a
+  | Just HRefl <- eqTypeRep (typeOf a) (typeRep @GHC.Msg) = case a of
```
hsyl20: It would be simpler to replace the explicit type equality check with `case fromDynamic (toDyn a) of ...`

asterius/src/Asterius/GHCi/Internals.hs
```diff
@@ -0,0 +1,415 @@
...
+    if GHC.failed ok
+      then GHC.throwGhcExceptionIO (GHC.ProgramError "")
+      else pure (pls, ())
+
+neededModules :: Data a => a -> [GHC.Module]
+neededModules = GHC.uniqDSetToList . w
+  where
+    w :: Data a => a -> GHC.UniqDSet GHC.Module
+    w t
+      | Just HRefl <- eqTypeRep (typeOf t) (typeRep @GHC.CoreBndr) =
```
hsyl20: Same

asterius/src/Asterius/GHCi/Internals.hs
```diff
@@ -0,0 +1,415 @@
...
+        (js_s, v) <-
+          maybe (fail "asteriusWriteIServ: RunTH: no JSSession") pure $
+            (\(x, _, v) -> (x, v))
+              <$> ghciJSSession s
+        asteriusRunModFinalizers hsc_env js_s v
+    GHC.Msg m -> fail $ "asteriusWriteIServ: unsupported message " <> show m
+  | otherwise = withMVar globalGHCiState $
+    \s -> let Just (_, p, _) = ghciJSSession s in writePipe p $ put a
+
+asteriusRunTH ::
```
hsyl20: Maybe use a datatype to pass all the arguments (and give them a name by doing so).

asterius/src/Asterius/GHCi/Internals.hs
```diff
@@ -0,0 +1,415 @@
...
+      evaluate s {ghciLibs = lib <> ghciLibs s}
+    GHC.LoadObj p -> modifyMVar_ globalGHCiState $ \s -> do
+      obj <- decodeFile p
+      evaluate s {ghciObjs = M.insert p obj $ ghciObjs s}
+    GHC.AddLibrarySearchPath _ -> pure $ GHC.RemotePtr 0
+    GHC.RemoveLibrarySearchPath _ -> pure True
+    GHC.ResolveObjs -> pure True
+    GHC.FindSystemLibrary lib -> pure $ Just lib
+    GHC.StartTH -> pure $ unsafeCoerce $ GHC.RemotePtr 0
+    _ -> fail "asteriusIservCall"
```
hsyl20: Can you add a case: `GHC.FreeHValueRefs _ -> pure ()`

asterius/src/Asterius/FrontendPlugin.hs
TerrorJack: `sPgm_i` is the path to the `iserv` executable. Since we use hooks to override all iserv logic, we don't need a standard iserv process, so we set it to `false` here; just in case ghc start `false` by accident, it'll exit by non-zero immediately to crash ghc, so we can spot where's iserv-related code which we forgot to override.

asterius/src/Asterius/GHCi/Internals.hs
TerrorJack: This makes sense. Will do shortly.

TerrorJack:
> `asteriusRunTH` and `asteriusHscCompileCoreExpr` could use more comments to indicate the different steps.

Makes sense. Will do an ASCII diagram in the comments later, to indicate what functions are called at which steps.

hsyl20:
I have reduced the error I get with this branch to:

```haskell
module Main where

main :: IO ()
main = return ()

{-# ANN module "HLint: ignore Use camelCase" #-}
{-# ANN module "HLint: ignore Reduce duplication" #-}
```

The failure is:
```
ahc: user error (hGet': expected 4 bytes, got 0)
^C
test/th/HLint.hs:6:1: error:
    • Exception when trying to run compile-time code:
        user interrupt
    • In the annotation:
        {-# ANN module "HLint: ignore Use camelCase" #-}
  |
6 | {-# ANN module "HLint: ignore Use camelCase" #-}
```

If we comment one of the two annotations, it works! There may be an issue when we have several splices in the same file?

TerrorJack:
@hsyl20 Multiple splices in the same module should work now.

TerrorJack:
There's still left work:

* A separate `JSSession` (and a `node` process) is set up with each splice, but `closeJSSession` is never called. Still unsure what's the best timing to finalize them; after handling `RunModFinalizers` message? Or after the next `StartTH` or `RunTH` message? Or just collect them and batch finalize upon `closeIServ`?
* Logging messages aren't cleaned up yet. The next TH wip branch will still have logging in the beginning anyway, so probably won't hurt to keep them for now.

Time for some real sleep I guess

TerrorJack:
The created `JSSession`s are now finalized when the global `JSSession` is updated, or when `closeIServ` is called. We don't assume `RunModFinalizers` is the safe place to finalize them; there may be future error recovery related messages when finalizers throw.

TerrorJack:
@hsyl20 Added source comments in the TH internal module. Do point out places where additional explanation is required!

hsyl20:
> @hsyl20 Added source comments in the TH internal module. Do point out places where additional explanation is required!

Awesome! I think you can merge.

Issue #268, opened 2019-09-10
Implement multi-threading


TerrorJack:
* [ ] Migrate some previously global state to thread-local, e.g. the context information of async FFI
* [ ] Modify the scheduler invocation method, instead of being directly invoked by `rts_eval*` calls, it starts as a "background daemon" async function, taking eval requests from a channel and writing back responses.
* [ ] Implement the run queue and modify the garbage collection invocation method, instead of using the current `StgTSO` object as the root object, use the run queue instead
* [ ] Implement proper blocking behavior for `MVar`s
* [ ] Implement the forking utilities
* [ ] Implement async exception utilities, e.g. `throwTo`
* [ ] Eyeball the concurrency related tests in `ghc-testsuite`

TerrorJack:
@hsyl20 Take a look at https://github.com/tweag/asterius/tree/wip-scheduler-daemon. This was the WIP branch of multi-threading, there`s a `Channel` implementation in the `rts` directory, and all `rts_eval*` calls just write to the `Channel`, which is picked up by the background scheduler.

TerrorJack:
And note that even when multi-threading is delivered, STM will be another pile of work and shall get a separate issue by then.

TerrorJack:
See https://gist.github.com/TerrorJack/a20a4512b31356e9c9a115bef08525d7 for an improved `Channel` implementation that ensures multiple synchronous `take` calls work. Also, the `wip-scheduler-daemon` branch is pretty outdated, you can rebase on current `master` before experimenting.

TerrorJack:
Closed by #268. Further improvements related to multi-threading should get individual issues and PRs.

Issue #269, opened 2019-09-10
Implement proper eventlogs


TerrorJack:
* [ ] Improve the underlying event log framework, allowing both textual and binary events, coming from both compiled Haskell code and the runtime itself
* [ ] Fix the `traceEvent#` and `traceBinaryEvent#` primops, and make sure the `-eventlog` ghc flag works
* [ ] Add more loggings in the runtime, ensuring there's minimal overhead when the event log isn't subscribed
* [ ] Eyeball the eventlog related tests in `ghc-testsuite`

PR #302, opened 2019-09-19
Implement basic multi-thread support

hsyl20:
This PR modifies the TSO scheduler to support multi-threading (cf #268).

`forkIO`, `threadDelay` and `myThreadId` work as expected (see added "rts" test).

Future work is needed to support more primitives (MVar, STM, etc.).


hsyl20:
@TerrorJack It is finally ready for review!

asterius/src/Asterius/Passes/SafeCCall.hs
```diff
@@ -158,36 +159,76 @@ splitSingleBlock vts save_instrs base_k base_block@RelooperBlock {..} =
...
+                   , offset = fromIntegral offset_StgTSO_ffi_return
+                   , ptr = Unary
+                       { unaryOp = WrapInt64
+                       , operand0 = getLVal currentTSO
+                       }
+                   , value = ConstI64 0
+                   , valueType = I64
+                   }
+               ]
+               ++ [ Store
```
hsyl20: I don't really understand why we need to reset to 0 all the saved regs after an FFI call, but if we don't bad things happen. Avoiding it or only reseting some of them would generate less code.

asterius/rts/rts.scheduler.mjs
```diff
@@ -0,0 +1,475 @@
...
+
+          // Add the thread into the run-queue
+          this.enqueueTSO(tso);
+
+          break;
+        }
+
+        case Cmd.yield: {
+          if (this.runQueue.length > 0) {
+            const tid = this.runQueue.shift();
```
TerrorJack: For now, it's fine for us to use an array for the run queue and `.shift()` for fetching the queue head. However, `.shift()` is usually O(n) (unlike `.pop()`) and can be a problem with a large number of runnable threads in the queue, so we also need to add a ticket about a more efficient queue implementation.

asterius/src/Asterius/Builtins.hs
```diff
@@ -117,19 +117,14 @@ rtsAsteriusModule opts =
           , ( "__asterius_func"
```
TerrorJack: `__asterius_func` used to be a global variable when multi-threads was absent. Now, if one thread calls an async import and suspends, and another thread does the same, `__asterius_func` of the first thread can get overwritten right? So this should also be made a thread-local thing like `__asterius_regs` and `__asterius_ret`

asterius/rts/rts.scheduler.mjs
```diff
@@ -0,0 +1,475 @@
...
+  returnFFIPromise(tid, ffiPromise) {
+    const tso_info = this.tsos.get(tid);
+    //console.log(`Thread ${tid}: return FFI Promise`);
+    tso_info.blockingPromise = ffiPromise;
+  }
+
+  /**
+   * Called when a thread stops for some reason.
+   */
+  returnFromTSO(tid) {
```
TerrorJack: Minor naming suggestion: `returnedFromTSO` is nicer since it's called *after* another round of cmm execution, but `returnFromTSO` sounds like it's used *during* execution to terminate it.

asterius/rts/rts.scheduler.mjs
```diff
@@ -0,0 +1,475 @@
...
+  /**
+   * Start the scheduler
+   */
+  run(exports) {
+    exports.context.reentrancyGuard.enter(0);
+    return this.scheduler_loop(exports).then
+      ( () => { exports.context.reentrancyGuard.exit(0); }
+      , e  => {
+          // signal all the TSOs that they won't complete
+          for (var [tid,tso_info] of this.tsos) {
```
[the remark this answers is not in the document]
hsyl20: I would keep this code. It's helpful to catch bugs in the scheduler when we work on it. Also maybe we could add a Terminate command to ask the scheduler to stop, in which case the reentrancy.exit is useful.

asterius/rts/rts.scheduler.mjs
```diff
@@ -0,0 +1,475 @@
...
+      );
+  }
+
+  /**
+   * Scheduler loop
+   */
+  async scheduler_loop(e) {
+    while (true) {
+      // read a command from the channel
+      this.running = false;
```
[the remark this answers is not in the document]
hsyl20: Not anymore.

asterius/rts/rts.scheduler.mjs
```diff
@@ -0,0 +1,475 @@
...
+  }
+
+  /**
+   * Called from a generated safe FFI import call.
+   *
+   * @param ffiPromise Promise executing the FFI import code asynchronously.
+   */
+  returnFFIPromise(tid, ffiPromise) {
+    const tso_info = this.tsos.get(tid);
+    //console.log(`Thread ${tid}: return FFI Promise`);
```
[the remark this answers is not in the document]
hsyl20: These have proved to be useful for debugging. I wondered if we could add a "trace" option to the scheduler instead of commenting them but it didn't want to inflict some overhead (probably minor, but anyway...).

ghc-toolkit/ghc-libdir/include/rts/storage/TSO.h
```diff
@@ -133,6 +133,10 @@ typedef struct StgTSO_ {
 
     struct StgTRecHeader_ * trec;       /* STM transaction record */
 
+    StgWord64               saved_regs[128]; /* Asterius: saved local regs */
```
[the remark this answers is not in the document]
hsyl20: I have checked and in this case it's fine as long as we add fields to the end of the StgTSO struct. GHC only uses some offsets as constants for this struct (phew!). It's better if we can avoid calling back into JS too often.

asterius/rts/rts.scheduler.mjs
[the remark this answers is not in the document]
TerrorJack: When #269 is implemented we shall use event logs instead of vanilla `console.log` calls, and event logs should be cheap when they're not actually enabled. But let's keep the `console.log` comments here for the time being.

ghc-toolkit/ghc-libdir/include/rts/storage/TSO.h
```diff
@@ -183,6 +183,19 @@ typedef struct StgTSO_ {
     StgWord32 saved_winerror;
 #endif
 
+    // Asterius fields must be at the end of the structure because the GHC
+    // compiler we get Cmm from isn't compiled with this header file but with
+    // the original one. Hence it assumes some offsets in the structure (see the
+    // generated constants by the deriveConstants program).
+    //
+    // Hopefully the size of TSO structure isn't a constant used by the
```
TerrorJack: To get confidence about how `StgTSO` sizes/offsets are used by the host ghc, grep for `StgTSO` in the [`platformConstants`](https://github.com/tweag/asterius/blob/master/ghc-toolkit/ghc-libdir/platformConstants) file. We're being lucky in this specific case.

hsyl20: Yes I did that. I chose the wrong word, I will s/Hopefully/Luckily/

asterius/src/Asterius/Passes/SafeCCall.hs
TerrorJack: We can call `memset` for smaller & faster reset code. The point of resetting is: a function with async imports may be entered multiple times, and we must ensure no state from previous async import calls is leaked.

asterius/rts/rts.scheduler.mjs
```diff
@@ -0,0 +1,473 @@
...
+                    tso_info.ffiRetType   = retTyp;
+                    // Store return block symbol
+                    tso_info.ffiAsyncFunc = this.memory.i64Load(tso + rtsConstants.offset_StgTSO_ffi_func);
+
+                    this.blockedTSOs.delete(tid);
+                    this.runQueue.push(tid);
+                    this.submitCmdWakeUp();
+                  }
+              , e => {
+                    tso_info.ffiRetErr = e;
```
TerrorJack: Why don't we re-enqueue the thread and submit a wakeup command in case of an FFI `Promise` rejection here?

hsyl20: Nice catch!

PR #304, opened 2019-10-01
Util scripts for cleaning & rebuilding

TerrorJack:
This PR adds two util scripts for convenient cleaning and rebuilding:

* `utils/coffee-break.sh` cleans up `ghc-toolkit`/`asterius`, rebuilds them, and executes `ahc-boot` with all CPU cores.
* `utils/noon-break.sh` does something similar, but disables parallelism for `ahc-boot` for deterministic ABI of boot cache, and flips on `ASTERIUS_DEBUG` so that IR dumps of boot libs are enabled, and Core/STG/Cmm linting is done. Takes longer time.

When rebooting is required, execute one of these scripts from the project root directory for proper cleaning and rebooting. Rebooting should be done when:

* The toolchain (underlying `ghc`) is updated
* Boot libs in `ghc-toolkit` is modified
* Wasm codegen logic in `asterius` which affects Cmm2Wasm is modified

### DIFF
--- a/utils/clean.sh
+++ b/utils/clean.sh
@@ -2,7 +2,9 @@
 
 export ROOT=$(stack path --local-install-root)
 
-stack clean asterius ghc-toolkit
+stack clean \
+  asterius \
+  ghc-toolkit
 rm -rf \
   $ROOT/share/*/asterius* \
   $ROOT/share/*/ghc-toolkit* \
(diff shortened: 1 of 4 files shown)


hsyl20:
Useful scripts! Can you add a `README` in `utils/` with the contents of this MR description?

PR #305, opened 2019-10-01
Implement MVar support

hsyl20:


ghc-toolkit/boot-libs/rts/PrimOps.cmm
```diff
@@ -1441,12 +1441,82 @@ stg_takeMVarzh ( P_ mvar /* :: MVar a */ )
      * and wait until we're woken up.
      */
     if (StgMVar_value(mvar) == stg_END_TSO_QUEUE_closure) {
-        jump stg_raisezh(base_GHCziIOziException_blockedIndefinitelyOnMVar_closure);
+        //if (info == stg_MVAR_CLEAN_info) {
```
TerrorJack: Why are we commenting out the `dirty_MVAR` logic here? Our gc currently doesn't use clean/dirty info of MutVars, MVars, etc, but dirtying it here won't hurt either. And we can have fewer hacks in this file.

ghc-toolkit/boot-libs/rts/PrimOps.cmm
```diff
@@ -1441,12 +1441,82 @@ stg_takeMVarzh ( P_ mvar /* :: MVar a */ )
...
+
+        SET_HDR(q, stg_MVAR_TSO_QUEUE_info, CCS_SYSTEM);
+        StgMVarTSOQueue_link(q) = END_TSO_QUEUE;
+        StgMVarTSOQueue_tso(q)  = CurrentTSO;
+
+        if (StgMVar_head(mvar) == stg_END_TSO_QUEUE_closure) {
+            StgMVar_head(mvar) = q;
+        } else {
+            StgMVarTSOQueue_link(StgMVar_tail(mvar)) = q;
+            //ccall recordClosureMutated(MyCapability() "ptr",
```
TerrorJack: Same question above; we can have a dummy `recordClosureMutated` builtin function so even less hacks here.

asterius/test/rts/MVar.hs
```diff
@@ -0,0 +1,45 @@
...
+import Control.Concurrent.MVar
+import Control.Monad
+import System.Mem
+
+foreign import javascript "console.log(${1})" js_print :: JSVal -> IO ()
+
+printString :: String -> IO ()
+printString s = js_print (coerce (toJSString s))
+
+main :: IO ()
```
TerrorJack: I suggest inserting `performGC` calls to make sure the `MVar` objects are properly handled by the gc.

asterius/rts/rts.scheduler.mjs
```diff
@@ -407,6 +410,18 @@ export class Scheduler {
     this.submitCmdWakeUp();
   }
 
+  /**
+   * Remove the given TSO from the "blocked" list, enqueue it in the run-queue
+   * and wake-up the scheduler
+   */
+  wakeupThread(tso) {
+    const tid = this.getTSOid(tso);
+    this.blockedTSOs.delete(tid);
```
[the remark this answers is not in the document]
hsyl20: Nope, I will remove it.

asterius/test/rts/MVar.hs
hsyl20: Currently the GC doesn't handle resurrected threads, hence we don't get any "indefinitely blocked on MVar exceptions". I'll add a ticket for this.

hsyl20: Done #310

asterius/src/Asterius/Boot.hs
```diff
@@ -108,6 +108,8 @@ bootRTSCmm BootArgs {..} =
                          "rts",
                          "-dcmm-lint",
                          "-O2",
+                         "-DASTERIUS",
+                         "-optc=-DASTERIUS",
```
TerrorJack: Only adding `-DASTERIUS` as a ghc flag is enough for it to be passed to the underlying CPP program. Do you think the `-DASTERIUS` should be unconditionally flipped on for `ahc`? Then we won't need to add this flag to a lot of places (here, boot script, patched cabal files in the future, etc)

ghc-toolkit/boot-libs/rts/PrimOps.cmm
```diff
@@ -1432,31 +1431,109 @@ stg_newMVarzh ()
...
+        StgMVarTSOQueue_tso(q)  = CurrentTSO;
+
+        if (StgMVar_head(mvar) == stg_END_TSO_QUEUE_closure) {
+            StgMVar_head(mvar) = q;
+        } else {
+            StgMVarTSOQueue_link(StgMVar_tail(mvar)) = q;
+            ccall recordClosureMutated(MyCapability() "ptr",
+                                             StgMVar_tail(mvar));
+        }
+        StgTSO__link(CurrentTSO)       = q;
```
TerrorJack: In our `StgTSO` object scavenge logic, did we handle the `__link` field as well?

hsyl20: Currently the GC evacuate all the TSOs (even the completed ones). I have updated #310 so that we don't forget to fix this.

TerrorJack: > evacuates all the TSOs
I think you really mean "scavenge" here? (Traversing all fields of an object), since TSOs are all pinned for now, we don't really evacuate them

hsyl20: Well we call "evacuateClosure" on them (the fact that they don't really move is an implementation detail IMO). But what I meant was that the `link` field contains a list of TSOs and we evacuate them all anyway, so it doesn't change anything if we consider this field or not during the TSO scavenging (for now).

PR #306, opened 2019-10-01
Update dependencies

TerrorJack:
This PR updates the project dependencies including:

* The stackage resolver, to latest LTS
* The ghc bindists, to the latest version including TH-related customizations
* The `inline-js` submodule, to the latest version containing a critical fix

It's required to remove the existing asterius ghc bindist in `~/.stack/programs/*`, otherwise `stack` won't pick up the new bindist and won't build later PRs. A `git clean` in the project directory is also recommended for a thorough cleanup.

PR #307, opened 2019-10-01
Update the ghci boot lib to include the prototype asterius ghci runner

TerrorJack:
This PR updates the `ghci` package to match the source code in our ghc fork, and also includes a prototypical implementation of ghci runner used by our current TH branch.

Note that #306 is the prequel of this PR; only merge before the prequel is merged.


hsyl20:
LGTM

PR #311, opened 2019-10-02
Implement PopCnt, Clz and Ctz for W8, W16 and W32

hsyl20:
Fix cgrun071 and cgrun077 in ghc-testsuite

### DIFF
--- a/asterius/src/Asterius/CodeGen.hs
+++ b/asterius/src/Asterius/CodeGen.hs
@@ -24,6 +24,7 @@ import Asterius.Passes.SafeCCall
 import Asterius.Resolve
 import Asterius.Types
 import Asterius.TypesConv
+import Asterius.EDSL
 import qualified CLabel as GHC
 import qualified Cmm as GHC
 import qualified CmmSwitch as GHC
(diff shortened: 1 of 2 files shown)


asterius/src/Asterius/CodeGen.hs
```diff
@@ -829,11 +832,31 @@ marshalCmmPrimCall (GHC.MO_Memcmp _) [_cres] [_ptr1, _ptr2, _n] = do
         }
     ]
 marshalCmmPrimCall (GHC.MO_PopCnt GHC.W64) [r] [x] =
-  marshalCmmUnPrimCall PopcntInt64 I64 r x
+  marshalCmmUnPrimCall PopcntInt64 I64 r x id
+marshalCmmPrimCall (GHC.MO_PopCnt GHC.W32) [r] [x] = do
```
TerrorJack: Is it possible to use `PopcntInt32` for the `W32` case? Same for `Clz` and `Ctz`.

hsyl20: I've done it but it seems to make the two tests fail again. I will try with a fresh build.

hsyl20: Ok fixed.

PR #312, opened 2019-10-02
Minor nomain runner improvement

TerrorJack:
This PR slightly improves the "nomain" runner we use to link and run a Haskell module without the default main closure:

* We no longer pass the raw content of the `.wasm` file through `inline-js-core`. Since the non-main linking is not fully done in-memory yet and we have the `.mjs`/`.wasm` files on disk anyway, we only send the path and use `fs` of `node` to load it.
* The `NoMain` module's exported buffer now performs some side effect (writing to `node` standard error). This is supposed to be used for testing "relinking", when the same closure is evaluated again, the side effect shouldn't occur for a second time because the CAF is already in WHNF.

The real support for "relinking" (appending new code onto existing in-memory asterius instance) takes much more tinkering and will come in separate PRs.

The diff noise is due to the switch to `ormolu` as the formatter everywhere.

### DIFF
--- a/asterius/src/Asterius/JSRun/NonMain.hs
+++ b/asterius/src/Asterius/JSRun/NonMain.hs
@@ -1,14 +1,27 @@
 {-# LANGUAGE OverloadedStrings #-}
 
 module Asterius.JSRun.NonMain
-  ( newAsteriusInstanceNonMain
-  ) where
+  ( distNonMain,
+    newAsteriusInstanceNonMain,
+  )
+where
 
-import Asterius.Ld (LinkTask(..), linkModules)
-import Asterius.Main (Target(..), Task(..), ahcDistMain)
+import Asterius.Ld
+  ( LinkTask (..),
+    linkModules,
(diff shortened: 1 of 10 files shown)


asterius/src/Asterius/JSRun/NonMain.hs
```diff
@@ -18,60 +31,62 @@ linkNonMain store_m extra_syms = (m, link_report)
...
+            linkLibs = [],
+            linkModule = mempty,
+            Asterius.Ld.debug = False,
+            Asterius.Ld.gcSections = True,
+            Asterius.Ld.binaryen = False,
+            Asterius.Ld.verboseErr = True,
+            Asterius.Ld.outputIR = Nothing,
+            rootSymbols = extra_syms,
+            Asterius.Ld.exportFunctions = []
           }
```
hsyl20: Sigh, "ormolu" doesn't seem to be configurable and I prefer the aligned-comma style :'(

Issue #319, opened 2019-10-10
Reflect our way implementation of `JSVal`s


TerrorJack:
Currently, we have `newtype JSVal = JSVal (StablePtr ())`, and use a tagging bit to identify regular `StablePtr`s and `JSVal`s. The garbage collector has special logic of handling `StablePtr` objects on the heap, and when `JSVal` objects are discovered, they are added to the "live" `JSVal` object set, and all non-live `JSVal` objects are freed from SPT after a round of gc.

The above design looks normal. The implementation seems to work. Except from occasional runtime whispers about `undefined` being fetched from SPT, probably a use-after-free situation for `JSVal`s?

That's right. And the design is **broken**. Fundamentally broken. For so long. The reason is simple: **unboxing**.

The `StablePtr#` primitive type is the real SPT index, and it's just an unsigned word without any guaranteed bit pattern. It can be boxed with a `StablePtr` closure, but it can also be stored from/to arbitrary addresses (there's a `Storable` instance for `StablePtr`), by then there won't be `StablePtr` closures, but the relevant SPT entries are still valid. Unboxing can also come from various GHC optimizations. So, the "live" `JSVal` object set cannot be obtained by simply looking for `StablePtr` closures, and we very likely introduced use-after-free of `JSVal`s (and consequently, `Integer`s) in the garbage collector!

Now that the problem is made clear. What are we supposed to do? At the very least, we must remove all `StablePtr`-closure related special handling in thecurrent gc, since we now see the pitfall. But how do we garbage collect `JSVal`s? Or is that even possible, under the current wasm feature set we're using?

* Simply give up the previous bluffing about garbage collection of `JSVal`s, tell users to finalize by hand.
* Conservative garbage collection of `JSVal`s: every non-pointer word scanned by gc is a potential `JSVal` SPT index.
* Use something which has the `UnliftedRep` runtime representation for the `JSVal` SPT index. They will be marked as pointers in the relevant info tables, and the gc will attempt to treat them as gc managed pointers. From there we can do a live `JSVal` set with confidence. It seems to be the GHCJS approach btw (GHCJS uses `ByteArray#`)

The last proposed approach seems to be the most practical method to fix our miserable `JSVal` story. But there's an extra cost: say that we do a `data JSVal = JSVal ByteArray#` as GHCJS does, then `JSVal` won't work as an FFI type. So there'll need extra work in the JSFFI processor.

TerrorJack:
Closed via #324.

PR #321, opened 2019-10-11
Fix JSVal implementation

TerrorJack:
Addresses #319


TerrorJack:
Roadmap:

* [x] Make typechecker/desugarer support `data JSVal`
* [ ] Move `JSVal` implementation module `Asterius.Types` to `ghc-prim`
* [ ] Base the current `Integer` definition on `JSVal`
* [ ] Modify JSFFI processor to work with new `JSVal`
* [ ] Modify RTS API to work with new `JSVal`

TerrorJack:
The new plan is using `StableName# ()` as the `JSVal` primitive representation. In the beginning I thought I'd choose `ByteArray#` like GHCJS does, but when `ByteArray#` is passed as a foreign import argument, GHC doesn't pass the (fake) `StgArrBytes` closure address, but adjusts it and passes the `payload` address instead. This works well for normal `ByteArray#` parameters but is an extra annoyance in our case. `StableName#` has an unlifted rep, passes FFI boundary without being modified, hence more suitable for this use.

TerrorJack:
All work done, but `Integer` seems to be broken for unknown reason. And it's a ~2k loc changeset making code review impractical, so will close this one for now. The code is still there and we'll split it into smaller PRs.

PR #323, opened 2019-10-12
`Any` and new `JSVal` as FFI types

TerrorJack:
This PR belongs to a series of PRs to split up #321; comes after #322 

This PR implements two changes:

* In `Asterius.Foreign.Internal`, we implement `isJSValTy`, which tells if a `GHC.Type` is `JSVal`. We match the `UnitId`, `ModuleName` and `OccName`. It checks the *future* version of `JSVal`, which should be defined in `Asterius.Prim` in `ghc-prim`. Changes to the current `JSVal` will go in future PRs.
* In `Asterius.Foreign`, we change our FFI typecheck/desugar hook code and support `Any` and future `JSVal` as FFI import/export argument/result types. `Any` will be used for passing closure addresses of boxed types.

### DIFF
--- a/asterius/src/Asterius/Foreign.hs
+++ b/asterius/src/Asterius/Foreign.hs
@@ -24,6 +24,7 @@ import ForeignCall
 import qualified GHC.LanguageExtensions as LangExt
 import GhcPlugins
 import HsSyn
+import MkId
 import OrdList
 import Outputable
 import Pair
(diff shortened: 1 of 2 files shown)


TerrorJack:
> Perhaps we should make a first PR that applies Ormolu to the whole codebase all at once?

Sounds good to me. Do you have any WIP branches that will be impacted by the format PR? If not I can do the formatting and merge back, before you proceed with reviewing.

hsyl20:
> > Perhaps we should make a first PR that applies Ormolu to the whole codebase all at once?
> 
> Sounds good to me. Do you have any WIP branches that will be impacted by the format PR?

I don't think so. And I can always rebase if necessary.

> If not I can do the formatting and merge back, before you proceed with reviewing.

Thanks. I will install Ormolu to avoid messing up with the format in later PRs.

PR #324, opened 2019-10-13
Migrate `JSVal` to the new definition

TerrorJack:
This PR belongs to a series of PRs to split up #321; comes after #323. Given the changeset is huge, better review before previous PRs are merged and `master` is merged back.

This PR implements the following changes:

* The origial `Asterius.Prim` in `base` is removed. It's not used anywhere.
* The original `Asterius.Types` in `base` is split up to two modules; most original code moves to `Asterius.Prim` in `ghc-prim`. `Asterius.Types` in `base` now only includes `JSException` related code, since `Exception`-related stuff only exists in `base`; to avoid breakage, it also exports `Asterius.Prim`.
* `Asterius.Prim` now implements the new `JSVal` definition, and `JSVal` is exported as an opaque datatype.
* `Asterius.Prim` now uses `Any` instead of `Addr#` for passing closure addresses.
* Some unused code in `Asterius.Magic` is pruned.
* We now implement `rts_mkJSVal` and `rts_getJSVal`, as a part of RTS API for working with the *lifted* `JSVal` closures and the underlying `JSVal` indices. The runtime also uses `rts_mkJSVal` instead of `rts_mkStablePtr` when creating `JSVal` closures.
* When parsing a `GHC.Type` to our own JSFFI type signature, we now only process the normalized type, since the previous trick which distincts between `JSVal` and regular `StablePtr` is no longer needed.
* In the garbage collector, we no longer scan for `StablePtr` closure fields to look for `JSVal` indices; instead, we check for the "closure address" passed to `evacuateClosure`, and if it doesn't has the higher 32-bit data address tag, we assume it's a `JSVal` index, and add it to the live `JSVal` set.

Note that the `Integer` type is still based on the old definition, with the old tagging trick for small `Integer`s. So the pitfall of use-after-free should be gone for `JSVal` but still present for `Integer`s. Changes to `Integer` shall go in the next PR.


asterius/src/Asterius/Builtins.hs
```diff
@@ -1239,6 +1260,31 @@ rtsMkHelper _ n con_sym = runEDSL n $ do
   storeI64 p 8 i
   emit p
 
+rtsMkBoolFunction _ = runEDSL "rts_mkBool" $ do
```
hsyl20: There are already the same functions below. Maybe a rebase issue?

TerrorJack: Major rebase issue that broke stuff :( Will ping again when fixed

PR #327, opened 2019-10-16
Install (and keep) netbase

### LINKED ISSUE #326: `stack build` fails in Docker container with "no such protocol name: tcp"
This is similar to https://github.com/snoyberg/http-client/issues/292.

```
$ docker run -it -v ~/mirror:/mirror terrorjack/asterius
root@fbd6f87dfab8:~/asterius# stack build                                                      
Exception while reading snapshot from lts-14.8 (8af5eb80734f02621d37e82cc0cde614af2ddc9c320610acb0b1b6d9ac162930,524789):
HttpExceptionRequest Request {                                                                                                                                                                 
  host                 = "raw.githubusercontent.com"                                                                                                                                           
  port                 = 443                                                                                                                                                                   
  secure               = True                                                                  
  requestHeaders       = [("User-Agent","Haskell pantry package")]                             
  path                 = "/commercialhaskell/stackage-snapshots/master/lts/14/8.yaml"                                                                                                          
  queryString          = ""                                                                    
  method               = "GET"                                                                 
  proxy                = Nothing                                                               
  rawBody              = False                                                                                                                                                                 
  redirectCount        = 10                                                                    
  responseTimeout      = ResponseTimeoutDefault                                                                                                                                                
  requestVersion       = HTTP/1.1                                                              
}                                                                                              
 (ConnectionFailure Network.BSD.getProtocolByName: does not exist (no such protocol name: tcp))
```

This works:
```
root@think5:~/asterius# apt-get update && apt-get install netbase
[...]
root@think5:~/asterius# stack build
[...]
```

My suggestion would be to install `netbase` in `Dockerfile` (and leave it installed).

curiousleo:
Fixes #326.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,6 +32,7 @@ RUN \
     libncurses-dev \
     libnuma-dev \
     make \
+    netbase \
     python-minimal \
     python3-minimal \
     xz-utils \


curiousleo:
Okay, so I realised after creating this PR that there is an assumption here. https://asterius.netlify.com/building.html prefixes the `docker` invocation example with
> Put input program in a directory (e.g. ~/mirror), then map the directory to a Docker volume:
So it looks like the purpose of the Docker image is to be able to _use_ the `asterius` compiler, as opposed to building `asterius` itself. I thought the Docker image was meant to be used for the latter too, hence this PR. But if the Docker image is _only_ meant to allow people to easily test out a built version of `asterius`, then this PR should not be merged.

TerrorJack:
See https://github.com/tweag/asterius/issues/326#issuecomment-542956109.

PR #329, opened 2019-10-20
Apply prettier to all JavaScript code in rts

TerrorJack:
This PR applies prettier to all JavaScript code in rts; future refactorings in the rts should be based on this work to avoid diff amplification. Also, we designate prettier as our official formatter for the JavaScript part.


curiousleo:
What do you think about adding a `prettier` build step to CI that fails if there is JS code not formatted with `prettier`? In my experience, this has done wonders to enforce correct formatting.

Similarly, if `ormolu` is the official Haskell formatter for `asterius`, perhaps that should be enforced in CI too?

I'm happy to create a ticket for this if at least one other person thinks this is a good idea.

hsyl20:
@curiousleo That would be good. I don't have an automated way to execute them locally yet, so checking commits before merging would be useful.

curiousleo:
- https://github.com/tweag/asterius/issues/333 for JS / `prettier`
- https://github.com/tweag/asterius/issues/334 for Haskell / `ormolu`

PR #338, opened 2019-11-18
Make the rts script self-contained

TerrorJack:
This PR makes the `rts.mjs` script self-contained; the same `newAsteriusInstance` exported function can be reused to initiate different asterius instances. This is useful in multiple cases:

* Multiple instances compiled from different `Main` modules in the browser
* Runtime TH linker


TerrorJack:
> Is there an easy way to generate the (now shared) rts.constants.mjs file?

The most obvious approach would be using a custom `Setup.hs` to put it in the Cabal datadir. But I'm reluctant to use more custom `Setup.hs` magic; they are a burden for porting stuff to nix.

In fact there are already a few things which are script-generated: `AutoApply.cmm`, `rts.unicode.mjs`, the `npm install`ed artifacts in `npm-utils`, and `BuildInfo` modules of `asterius` and `ghc-toolkit`. We should really seek a uniform way to properly generate these things, but that shall go in a separate issue imo

hsyl20:
> In fact there are already a few things which are script-generated

Yes but do we have a script to generate the constant file? It would only contain something like:

```haskell
main = builderWriteFile out_rts_constants rtsConstants
```
which is removed in this PR.

> but that shall go in a separate issue imo

Agreed. As long as we have an issue tracking this, it is fine to not have this automated.

PR #340, opened 2019-11-20
Fix & test ahc-cabal

TerrorJack:
Whatever not tested on CI should be assumed broken. And at least not the case anymore for `ahc-cabal`..

Changes:

* Previously, `ahc-cabal` works by intercepting command-line arguments, adding a bunch of asterius-specific stuff and passing to `cabal`; it requires quite a bit of knowledge about how `cabal` arguments are parsed, and only work for a certain set of "common" commands. Now we take an alternative approach: obtain the cabal config file path, create and supply a temporary config file which is patched with `cabal user-config update`. This is simpler and works for all cabal commands.
* The frontend plugin of `ahc` now always hard-code the `ahc-ld` as the linker, so no need for passing `-pgml` from the command line
* The `ahc-ld` executable used to expect a `--prog-name` switch which is only generated by `ahc-link`, not `ahc` itself. We now default to an empty program name when the switch is absent.
* On CI we test a minimal `ahc-cabal` case: build and run `hello` from hackage.
* `cabal` is installed in the docker image. So `ahc-cabal` should work in the image.

TLDR: these problems are fixed:

* `ahc` fails to build executable due to the `ahc-ld` bug
* `ahc-cabal` fails to build executable targets due to the bug above
* `ahc-cabal` doesn't work for V1 build commands
* `ahc-cabal` doesn't work in the docker image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,13 +41,14 @@ jobs:
             apt install -y nodejs
             mkdir -p /root/.local/bin
             curl -L https://get.haskellstack.org/stable/linux-x86_64.tar.gz | tar xz --wildcards --strip-components=1 -C /root/.local/bin '*/stack'
+            curl -L https://downloads.haskell.org/~cabal/cabal-install-latest/cabal-install-3.0.0.0-x86_64-unknown-linux.tar.xz | tar xJ -C /root/.local/bin 'cabal'
       - checkout
 
       - run:
           name: Boot
           command: |
             git submodule update --init --recursive
-            stack --no-terminal -j2 build --test --no-run-tests
+            stack --no-terminal -j2 install --test --no-run-tests asterius
             stack --no-terminal exec ahc-boot
 
       - persist_to_workspace:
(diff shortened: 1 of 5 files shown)


TerrorJack:
Side note: `ahc-cabal` shouldn't be invoked with `stack exec`, due to the extra environment variables. However without the added stuff in `PATH`, it's likely that stuff like `hsc2hs` can't be found. Should we hard-code path of all tools in `ahc`? Or should we implement the hack to wipe out the `GHC*` and `HASKELL*` environment variables in `ahc-cabal`? Or is this a non-problem at all? Worth digging further but since basic stuff seems to be working, this should be enqueued as a lower priority thing to look at..

TerrorJack:
Another side note: currently `getProgName` should return an empty string when it's compiled by `ahc` (and transitively `ahc-cabal`). `ahc-link` is able to figure out a correct "program name" from `--input-hs`, but making `ahc` aware of the program name when performing linking is quite some more work.

Issue #342, opened 2019-11-26
Handle packages with custom `Setup.hs`


TerrorJack:
Issue discovered when stress testing TH stuff with `ahc-cabal new-install singletons`. For packages with custom `Setup.hs` (like `singletons`), the `Setup.hs` script gets compiled by `ahc`, and the resulting pseudo-executable can't be executed like regular ELF executables. So the subsequent steps all fail.

PR #344, opened 2019-11-26
Workaround lack of custom `Setup.hs` support

TerrorJack:
Related: #342 

For now, producing an ELF "executable" with wasm stuff is too much work, especially given the constraint of not patching Cabal. Here we do a workaround: when we detect the output path is something like `../dist/setup/setup`, we short-circuit `ahc` logic and call `ghc` to produce the executable, with the original ghc global pkgdb.

My eyes are bleeding with this hack. But at least it works to make `singletons` properly (?) pass the configure step and kick off building. And maybe for other packages out there as well, iff their custom `Setup.hs` doesn't have any `setup-deps` outside the ghc global pkgdb.

### DIFF
--- a/ghc-toolkit/src/Language/Haskell/GHC/Toolkit/FakeGHC.hs
+++ b/ghc-toolkit/src/Language/Haskell/GHC/Toolkit/FakeGHC.hs
@@ -8,12 +8,14 @@ module Language.Haskell.GHC.Toolkit.FakeGHC
 where
 
 import Control.Monad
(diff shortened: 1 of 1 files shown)


hsyl20:
Can you add comments describing the gory hack in the code too please? So that we don't forget to fix it properly in the future.

On the hack itself, I guess it's the price to pay to get cross-compilation working with stock Cabal for now :/

TerrorJack:
@hsyl20 Related comments are added.

Issue #347, opened 2019-11-27
Builtin `aeson` support


TerrorJack:
In the existing haskell-to-web use cases, `aeson` is still the dominant standard for marshaling Haskell data structures to JavaScript (and vice versa). We shall support `aeson` out-of-the-box:

* [x] Install and ship `aeson` into the global pkgdb using `ahc-cabal`
* [x] Base `asterius-prelude` on top of `aeson`
* [x] Implement runtime functions related to UTF-8 JSON encoding/decoding
* [x] Examples and tests

therewillbecode:
Watching this with interest.

TerrorJack:
Using only standard `JSON`, `TextDecoder` and `TextEncoder` APIs, I can come up with roughly two different approaches for marshaling:

## Through strict `ByteString`s

This approach marshals Haskell values first to lazy `ByteString`s, then strict ones, then decode the UTF-8 buffer to a JS string, finally to a JS value, likewise for the other direction.

I'm concerned about the performance of this approach, a lot of buffer copying happens here. There are tricks to reduce one or two copying operations:

* When passing strict `ByteString` to the `TextDecoder`, we expose the linear memory pointer directly as a typed array.
* Before calling `TextEncoder`, we allocate the `ByteString` buffer and write into it with `encodeInto`.

## Through `Value`s

This approach marshals Haskell values to a `Value` directly, then build the JS value by traversing the `Value`. For the other direction, the `Value` is built directly either via a runtime function or a Haskell function using JSFFI.

Given we have fast `Text` and JS string codec, this one may be faster. But of course, a lot of extra trouble. And not using vanilla `JSON.parse` & `JSON.stringify` may be a correctness risk?

@hsyl20 Any thoughts which route to take?
@luite Can you provide some insight on how ghcjs accomplishes the conversion? Does it need to go through `ArrayBuffer` and JS strings?

TerrorJack:
We implemented the codec between lazy `ByteString`s and `JSString`s, and decided to use `JSON.parse` & `JSON.stringify` on the js side. Directly working with `Value` closures on the heap seems might be a premature optimization for now.

TerrorJack:
Closed via #363. May revisit in the future when proved to be a bottleneck.

PR #348, opened 2019-11-27
Include `aeson` and friends into the global package database

TerrorJack:
First steps of #347 

This PR uses `ahc-cabal` to install `aeson` and friends into the global package database during booting. As a consequence, `ahc-link` or vanilla `ahc` should be able to compile main modules which use `aeson` stuff now. We also add `aeson` as a dependency of `asterius-prelude`, similar to how `ghcjs-base` depends on `aeson`.

It's worth noting that we now have two mixed mechanisms for building non-`rts` Cabal packages during booting: one is directly calling the `Setup` executable and using the in-tree package sources; the other is using `ahc-cabal` and downloading from Hackage. We should gradually migrate to use `ahc-cabal` as the primary driver, but that's extra work in the future.

@hsyl20 What packages other than `aeson` should we ship? This PR is a good timing to add them.


hsyl20:
> @hsyl20 What packages other than aeson should we ship? This PR is a good timing to add them.

`diagrams` ;-)

Maybe we shouldn't slow down the booting process too much but adding too many packages. However `lucid` would be a good candidate.

TerrorJack:
> Maybe we shouldn't slow down the booting process too much but adding too many packages.

I agree. `aeson` takes just a few more minutes on my own thinkpad so I think we're good for now.

> However lucid would be a good candidate.

It may be. But I think `lucid` should belong to an optional package set which is installed outside the main booting process; this is the "asterius platform" (similar to haskell platform) thing I've thought some time ago, a subset of recent stackage lts snapshot which is pre-installed in the global pkgdb of our pre-built docker image.

I'll be merging this one for now if no objections? :)

Issue #354, opened 2019-11-28
Tracking issue for known to compile packages


TerrorJack:
This is a tracking issue to maintain a list for known to compile packages, from a recent stackage lts snapshot (`lts-14.16`). Naturally, compiling != working, but getting stuff to compile is a good first step anyway.

```
Boolean-0.2.4
Cabal-2.4.1.0
Diff-0.3.4
Glob-0.10.0
HUnit-1.6.0.0
HsYAML-0.1.2.0
ListLike-4.6.2
MemoTrie-0.6.9
MonadRandom-0.5.1.1
NumInstances-1.4
Only-0.1
QuickCheck-2.13.2
SHA-1.6.4.4
StateVar-1.2
abstract-deque-0.3
abstract-par-0.3.3
aeson-1.4.6.0
aeson-compat-0.3.9
aeson-pretty-0.8.8
ansi-terminal-0.9.1
ansi-wl-pprint-0.6.9
appar-0.1.8
array-0.5.3.0
asterius-prelude-0.0.1
async-2.2.2
attoparsec-0.13.2.3
attoparsec-iso8601-1.0.1.0
auto-update-0.1.6
base-4.12.0.0
base-compat-0.10.5
base-compat-batteries-0.10.5
base-orphans-0.8.1
base64-bytestring-1.0.0.2
basic-prelude-0.7.0
binary-0.8.6.0
binary-orphans-1.0.1
bitarray-0.0.1.1
blaze-builder-0.4.1.0
blaze-html-0.9.1.2
blaze-markup-0.8.2.3
bsb-http-chunked-0.0.0.4
byteable-0.1.1
byteorder-1.0.4
bytestring-0.10.8.2
cabal-doctest-1.0.8
call-stack-0.1.0
case-insensitive-1.2.0.11
cassava-0.5.2.0
cborg-0.2.2.0
cborg-json-0.2.1.0
cereal-0.5.8.1
charset-0.3.7.1
chunked-data-0.3.1
circle-packing-0.1.0.6
cmdargs-0.10.20
colour-2.3.5
conduit-1.3.1.1
constraints-0.10.1
containers-0.6.0.1
contravariant-1.5.2
cookie-0.4.5
criterion-measurement-0.1.2.0
crypto-api-0.13.3
css-text-0.1.3.0
data-default-0.7.1.1
data-default-class-0.1.2.0
data-default-instances-containers-0.0.1
data-default-instances-dlist-0.0.1
[...]

PR #355, opened 2019-11-29
Ship a subset of Stackage LTS packages in the pre-built Docker image

TerrorJack:
This PR installs additional packages in the global pkgdb of the pre-built Docker image, so the packages can be used directly by `ahc-link`/`ahc`. Changeset:

* (Temporary) Remove the `master` branch restriction when building the image on CI
* Checks in a `cabal.config` file which is provided by the Stackage server, pinning exact versions for most packages
* Include a `lts.sh` build script to be invoked after booting when building the image

The [`lts.sh`](https://github.com/tweag/asterius/blob/wip-asterius-platform/lts.sh) file contains the list of packages to build. They are picked from the complete [list](https://github.com/tweag/asterius/issues/354) of known-to-compile packages. The principle of picking packages here should be:

* Actively maintained and belong to recent stackage lts snapshots and
* Not making the total compile time of the image on CI exceed memory or time limit

For brevity of the build script we don't put every package name into the build command; the packages listed are relatively high up in the dependency graph, so to bring in a lot of other common packages.

We can still add more packages here imo, as long as the CI job runs fine. It doesn't impact the rest of the pipeline anyway.


TerrorJack:
> It seems to fail on CI

I need to snipe some packages out of the list, will merge once the image builds on CI :)

Issue #357, opened 2019-12-02
Missing rts cbits for `text`


TerrorJack:
`Data.Text.Encoding` imports these functions for encoding/decoding:

* `_hs_text_decode_utf8`
* `_hs_text_decode_utf8_state`
* `_hs_text_decode_latin1`
* `_hs_text_encode_utf8`

This issue is blocking #347 #328 #257

TerrorJack:
The easiest thing to do would be patching `Data.Text.Encoding` and avoiding ccalls at all. But let's see if we can fix vanilla `text` without too much trouble.

TerrorJack:
Implemented and tested `_hs_text_decode_utf8` and `_hs_text_encode_utf8`, sufficient to unblock #347. Will leave this one open for some more time since the cbits stuff still needs more work.

TerrorJack:
This still blocks #315:

```
file:///home/admin/asterius/asterius/test/diagrams/SquareLimit.mjs:4
process.on("unhandledRejection", err => { throw err; });
                                          ^

RuntimeError: RuntimeError: barf: _hs_text_decode_utf8_state
    at ExceptionHelper.barf (file:///home/admin/asterius/asterius/test/diagrams/rts.exception.mjs:143:13)
    at barf (wasm-function[7998]:0x2e7077)
    at textzm1zi2zi3zi1zmLZZOn9keu2Vl3XvFWCbVXmI_DataziTextziEncoding_.Lc2lfX (wasm-function[14173]:0x51d03d)
    at scheduleTSO (wasm-function[13906]:0x502579)
    at scheduleTSO_wrapper (wasm-function[13907]:0x5025c9)
    at Scheduler.scheduler_loop (file:///home/admin/asterius/asterius/test/diagrams/rts.scheduler.mjs:424:15)
    at file:///home/admin/asterius/asterius/test/diagrams/rts.scheduler.mjs:298:15
ahc-link: callProcess: node "--experimental-modules" "SquareLimit.mjs" (exit 1): failed
diagrams: callProcess: ahc-link "--input-hs" "test/diagrams/SquareLimit.hs" "--run" "--verbose-err" "--binaryen" (exit 1): failed
```

PR #360, opened 2019-12-03
Implement missing utf8 codec cbits for text

TerrorJack:
Related: #357 #347 

This PR uses W3C `TextDecoder` and `TextEncoder` to implement UTF-8 codec functions: `_hs_text_decode_utf8` and `_hs_text_encode_utf8`, which are required by `Data.Text.Encoding` and `aeson` related code.

The behavior of `_hs_text_decode_utf8` diverges with original c version with illegal input; in which case a runtime error is thrown in js and can't be handled in hs. It's possible to implement correct "stop at first illegal code point" behavior by bisecting the source buffer and feeding to `TextDecoder`, but IMHO that's more work and should be fixed in future PRs.

### DIFF
--- a/asterius/rts/rts.text.mjs
+++ b/asterius/rts/rts.text.mjs
@@ -23,4 +23,35 @@ export class TextCBits {
     }
     return 0;
   }
+
+  _hs_text_decode_utf8(dest, destoffp, src, srcend) {
+    const dec = new TextDecoder("utf-8", { fatal: true }),
+      s = dec.decode(
+        this.memory.i8View.subarray(Memory.unTag(src), Memory.unTag(srcend))
+      );
+    for (let i = 0; i < s.length; ++i)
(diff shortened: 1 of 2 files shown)


asterius/rts/rts.text.mjs
```diff
@@ -23,4 +23,35 @@ export class TextCBits {
...
+          Memory.unTag(src + srcoff * 2 + srclen * 2)
+        )
+      ),
+      dest = Number(this.memory.i64Load(destp)),
+      enc = new TextEncoder(),
+      l = enc.encodeInto(
+        s,
+        this.memory.i8View.subarray(
+          Memory.unTag(dest),
+          Memory.unTag(dest + srclen * 3)
```
hsyl20: Can you add the comment `see https://github.com/haskell/text/issues/194 for why srclen*3 is enough` here. I had to look for it to understand

Issue #358, opened 2019-12-02
Missing FNV hashing cbits for `hashable`


TerrorJack:
As the title says. This makes `unordered-containers` and `aeson` broken for now. Since @bollu implemented FNV hashing for our `Typeable`-related MD5 stub code some time ago, maybe we can seek to move that out from `rts.md5.mjs` and expose two sets of interfaces, one for `base`, another for `hashable`.

This issue blocks #347

TerrorJack:
For reference: https://github.com/ghcjs/ghcjs/blob/ghc-8.6/lib/boot/shims/pkg/hashable.js

TerrorJack:
Closed via #361.

PR #361, opened 2019-12-03
Migrate 128-bit FNV-1a logic to be used by base and hashable

TerrorJack:
Related: #358 #146 

We already have 128-bit FNV-1a algorithm implemented in `rts.md5.mjs` as pseudo-MD5 to be used by `GHC.Fingerprint` logic. This PR factors out the hashing logic, and implements interface for both `GHC.Fingerprint` logic and `hashable` logic. `hashable` uses 64-bit hashing, so the resulting 128-bit result shall be truncated before returning.

Side note: maybe we should also use `arithmoi` to write a script for a principled way of generating magic numbers for FNV hashing.

### DIFF
--- a/asterius/rts/rts.fnv.mjs
+++ b/asterius/rts/rts.fnv.mjs
@@ -1,0 +1,30 @@
+// We implement the 128-bit FNV-1a hash function:
+// https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function
+// Original work by @bollu. This module is used in two cases:
+// 1. Pseudo-MD5 hashing for Fingerprints in base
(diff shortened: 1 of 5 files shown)


TerrorJack:
So for the FNV code in `hashable`, in the end we decided to implement it as pure wasm functions in our EDSL. Since reusing the js implementation in `rts.fnv.mjs` brings in the additional problem of i64 FFI (can't reliably pass hash values).

There's another blob of c code in `hashable`: `hashable_getRandomBytes`. It's used by `Data.Hashable.RandomSource`, but I don't find any usages of the module inside `hashable` itself. So leaving that alone for now.

We didn't add tests for this PR; it'll prove to work in later `aeson` related PR soon.

PR #363, opened 2019-12-04
Implement `aeson` conversion interface

TerrorJack:
Addresses #347. To be reviewed after #361 is merged and this one is rebased.

This PR implements `Asterius.Aeson` providing:

```haskell
jsonToJSVal :: ToJSON a => a -> JSVal
jsonFromJSVal :: FromJSON a => JSVal -> Either String a
```

This enables marshaling between Haskell datatypes with `FromJSON` & `ToJSON` instances with JavaScript values. It calls `JSON.stringify` & `JSON.parse` under the hood, and so we also need to implement marshaling between UTF-8 encoded lazy `ByteString`s and UTF-16 encoded `JSString`s. 

This is enough to get stuff working for now, but `aeson` marshaling logic looks like a hot spot in typical applications so in the future we may still need some dark arts to speed things up.

### DIFF
--- a/ghc-toolkit/boot-libs/asterius-prelude/src/Asterius/UTF8.hs
+++ b/ghc-toolkit/boot-libs/asterius-prelude/src/Asterius/UTF8.hs
@@ -1,0 +1,48 @@
+module Asterius.UTF8
+  ( utf8ToJSString,
+    utf8FromJSString,
+  )
+where
+
+import Asterius.Magic
+import Asterius.Prim
+import qualified Data.ByteString.Internal as BS
+import qualified Data.ByteString.Lazy as LBS
+import qualified Data.ByteString.Unsafe as BS
+import Foreign
+
+{-# INLINE utf8ToJSString #-}
+utf8ToJSString :: LBS.ByteString -> JSString
+utf8ToJSString s
+  | LBS.null s = js_str_empty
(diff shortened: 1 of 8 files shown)


ghc-toolkit/boot-libs/asterius-prelude/src/Asterius/UTF8.hs
```diff
@@ -0,0 +1,48 @@
...
+{-# INLINE utf8FromJSString #-}
+utf8FromJSString :: JSString -> LBS.ByteString
+utf8FromJSString s = accursedUnutterablePerformIO $ do
+  l <- js_str_len s
+  if l == 0
+    then pure mempty
+    else fmap LBS.fromStrict $ BS.createUptoN (l * 3) $ \p ->
+      js_utf8_from_str s p (l * 3)
+
+foreign import javascript "''" js_str_empty :: JSString
```
hsyl20: Do we really have to `foreign import` this? Don't we have a way to create a JSString more directly?

TerrorJack: To get a `JSVal` in the Haskell heap we need to call a `foreign import javascript` anyway, I don't think there's a more direct approach.

PR #372, opened 2019-12-05
Fix Integer & JSVal garbage collection

TerrorJack:
Fixes #349. Before the shiny new Integer lands, we use `StableName# ()` as the `Integer` representation, so GHC would properly treat it as a boxed pointer, so `evacuateClosure` can be properly called on them.


TerrorJack:
Major regression shown by `asterius-test`. Need to investigate.

Issue #374, opened 2019-12-05
Garbage collector is very slow


hsyl20:
Currently the garbage collector is very slow. As we can see in the following flame graph (diagrams' SquareLimit example running in Chromium), we spend most of the time in the garbage collector and only a small fraction of the time doing actual work (in pink on the left of "returnFromTSO").

![screenshot-20191205@174056](https://user-images.githubusercontent.com/96633/70255469-24952780-1787-11ea-99c3-98881a20fba3.png)

TerrorJack:
A short-term mitigation we can easily implement: do not always enter gc when `performGC()` is entered. We can add a rts parameter specifying a threshold, and when the linear memory doesn't grow beyond the threshold, `performGC()` returns immediately and the rts only allocates fresh mblocks. We can also implement "idle gc" which performs gc when the TSO run queue is empty.

In the longer term, we should move away from naive copying collection to another algorithm.

hsyl20:
PR #375 reduces the GC time on this example from ~150ms to ~65ms.

@TerrorJack  We should try your approach to get even better perf

TerrorJack:
Should have been closed by #379.

PR #375, opened 2019-12-05
Enhance performance by avoiding BigInt when possible

hsyl20:
Reduce GC duration from ~150ms to ~65ms in SquareLimit example (cf #374)


TerrorJack:
Regression in debug mode: https://circleci.com/gh/tweag/asterius/7985

PR #378, opened 2019-12-06
Implement "--nursery-size" option for ahc-link & ahc-dist

TerrorJack:
Implements a part of GC mitigations discussed in #374.

We now have a `--nursery-size=` option for `ahc-link` and `ahc-dist`. It specifies the number of mblocks (for now, 1 mblock == 1mb) to be allocated when `updateNursery()` is invoked after heap overflow. The previous behavior is 1 mblock in most cases, we now set the default number to 64, to reduce chances of heap overflow and gc invocation.

It's worth noting that the initial nursery allocated in `hs_init` is unaffected and is still 1 mblock, which means even if the specified nursery size is big enough, at least 1 gc invocation is still required. Correcting  the initial nursery size would require the option to be passed to `ahc-ld`, which doesn't play nicely with `ahc-cabal` users.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,6 +127,7 @@ jobs:
             stack --no-terminal test asterius:sizeof_md5context
             stack --no-terminal test asterius:largenum
             stack --no-terminal test asterius:bytearray --test-arguments="--yolo"
+            stack --no-terminal test asterius:bytearray --test-arguments="--nursery-size=128"
             stack --no-terminal test asterius:fib --test-arguments="--no-gc-sections"
             stack --no-terminal test asterius:fib --test-arguments="--binaryen --no-gc-sections"
 
--- a/asterius/rts/rts.gc.mjs
+++ b/asterius/rts/rts.gc.mjs
@@ -20,7 +20,8 @@ export class GC {
     export_stableptrs,
     symbol_table,
     reentrancy_guard,
-    yolo
+    yolo,
+    nursery_size
   ) {
     this.memory = memory;
     this.heapAlloc = heapalloc;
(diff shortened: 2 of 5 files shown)


TerrorJack:
Regression discovered by the diagrams example:

```
RuntimeError: RuntimeError: Invalid live mblock 0x1ffff701200100
    at HeapAlloc.handleLiveness (file:///home/admin/asterius/asterius/test/diagrams/rts.heapalloc.mjs:102:15)
    at GC.performGC (file:///home/admin/asterius/asterius/test/diagrams/rts.gc.mjs:741:20)
    at Scheduler.returnedFromTSO (file:///home/admin/asterius/asterius/test/diagrams/rts.scheduler.mjs:115:17)
    at Scheduler.scheduler_loop (file:///home/admin/asterius/asterius/test/diagrams/rts.scheduler.mjs:425:18)
    at file:///home/admin/asterius/asterius/test/diagrams/rts.scheduler.mjs:298:15
```

Most likely caused by the incorrect `bdescr()` function in `rts.gc.mjs` which is used to calculate the nursery's `bdescr` address given a closure address. The previous assumption was: closure nurseries were always 1mb sized, and the assumption has been invalidated. Need to fix and test this before merging.

TerrorJack:
Proper `bdescr()` requires querying the `HeapAlloc` internal state (which is updated throughout the whole gc process). We need a bitwise trie or BST to speed it up since `bdescr()` is a hotspot in gc logic.

But maybe changing the nursery size is a wrong direction. We really should keep 1 mblock sized nurseries, and `bdescr()` would still work via a simple bit operation. We can skip GC occasionally though.

PR #381, opened 2019-12-07
Hack to support kmettverse packages

TerrorJack:
This PR doubles the dose of hack introduced in #344. When we attempt to compile the custom `Setup.hs` script with the host ghc, it may fail (e.g. lacking certain `custom-deps`), in which case we resort to compiling & using a `Setup.hs` that implements `build-type: Simple` behavior.

This hack is even grosser, but it fixes the building of the kmettverse packages (up to `lens` at least). Because kmettverse packages has a constant pattern of using `Setup.hs` with `cabal-doctest` for doctests, and since we aren't running tests anyway, the packages build fine with the default `build-type: Simple` script.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,7 @@ jobs:
             apt install -y nodejs
             mkdir -p /root/.local/bin
             curl -L https://get.haskellstack.org/stable/linux-x86_64.tar.gz | tar xz --wildcards --strip-components=1 -C /root/.local/bin '*/stack'
-            curl -L https://downloads.haskell.org/~cabal/cabal-install-latest/cabal-install-3.0.0.0-x86_64-unknown-linux.tar.xz | tar xJ -C /root/.local/bin 'cabal'
+            curl -L https://downloads.haskell.org/~cabal/cabal-install-2.4.1.0/cabal-install-2.4.1.0-x86_64-unknown-linux.tar.xz | tar xJ -C ~/.local/bin 'cabal'
       - checkout
 
       - run:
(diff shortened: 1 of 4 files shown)


TerrorJack:
Confirmed to work locally with `v1-install`, but `v2-install` fails with:

```
Failed to build distributive-0.6.1. The failure occurred during the configure
step.
Build log (
/root/.cabal/logs/ghc-8.6.5/distributive-0.6.1-71b08f7a52a6f7aa7c7b32fca44c47298541549566d84ca26817ee583a172aef.log
):
[1 of 1] Compiling Main             ( /tmp/cabal-install.-4825/dist-newstyle/tmp/src-4825/distributive-0.6.1/dist/setup/setup.hs, /tmp/cabal-install.-4825/dist-newstyle/tmp/src-4825/distributive-0.6.1/dist/setup/setup.o )
[1 of 1] Compiling Main             ( /tmp/cabal-install.-4825/dist-newstyle/tmp/src-4825/distributive-0.6.1/dist/setup/setup.hs, /tmp/cabal-install.-4825/dist-newstyle/tmp/src-4825/distributive-0.6.1/dist/setup/setup.o )
Linking /tmp/cabal-install.-4825/dist-newstyle/tmp/src-4825/distributive-0.6.1/dist/setup/setup ...
unrecognized 'configure' option `--disable-executable-static'
cabal: Failed to build distributive-0.6.1 (which is required by lens-4.18.1).
See the build log above for details.
```

Quick guess: the custom `Setup.hs` is compiled with the host `ghc-8.6.5` and `Cabal-2.4.0.1`, but the resulting `Setup` executable is assumed to be compiled with `Cabal-3.0.0.0`, and thus, mismatches in the accepted command line argument set.

TerrorJack:
Wasted an afternoon trying to backport `Cabal-3.0.0.0` back to `ghc-8.6`. But actually, downgrading `cabal-install` to `2.4.1.0` is enough to make our hack work for `v2-install` as well.

The problem with `cabal-install-2.4.1.0` is that `--installdir` isn't available yet. Need to find a way to locate the binary in the cabal store before downgrading.

PR #386, opened 2019-12-10
Use stock `integer-simple` and recover `JSVal` garbage collection

### LINKED ISSUE #385: Replace `BigInt`-backed `integer-simple` with the stock version
About the same time last year, we introduced a `BigInt` backed `integer-simple` package, hoping that the performance of platform-native `BigInt` is satisfactory, and simplifies the implementation of all operations. Revisiting the decision one year later, it seems to do more harm than good; we had the hidden use-after-free issue for so long, plus very rare but hard to debug `float unrepresentable in integer range` problem, and above all, the performance benefits are dubious (e.g. see #375).

Before the experimental new `integer-gmp` package checks in, we should phase out the current `integer-simple`, use the stock version, implement missing rts cbits related to floating-point decompositions, re-enable `JSVal` garbage collection and ensure there's zero regression in `ghc-testsuite`.

TerrorJack:
Closes #385 

There are regressions now: `decodeFloat` in `GHC.Float` is broken, as seen in the `fib` test. There are also 4 regressed tests in `ghc-testsuite` right now. We are unsure about impact on downstream packages.

Debugging floating-point bit operations is not pleasant. Will spend some more time to see if the existing floating-point cbits can be replaced with slower but simpler implementation.


TerrorJack:
It finally works! We included a `scalbn` implementation by reading the dump of `scalbn.o` compiled by `wasi-sdk` and writing it by hand. It's sufficient to fix the `__word_encodeDouble` functions. The real proper way would be integrating `wasi-sdk` and seek a way to reuse the `libc` generated there, but that should be left as future work.

TerrorJack:
This PR now only contains the logic of:

* Replacing existing `BigInt`-based `integer-simple` with the stock version. The `integer-simple` contents are identical with the ghc tree so no need to review this part
* Removing most of existing integer manager logic since they are no longer used
* Recover `JSVal` gc logic

It's suitable for review now.

Issue #388, opened 2019-12-11
Implement and use `wasm-opt` interface in `binaryen` binding


TerrorJack:
Besides its C API, `binaryen` provides several standalone utilities, one of which is `wasm-opt`, which implements end-to-end optimization passes for wasm binary modules. It is useful to both the `wasm-toolkit` and `binaryen` backend.

We should:

* [ ] Implement simple bindings that map to the `wasm-opt` command-line interface and call it
* [ ] Implement `ahc-link` & `ahc-dist` link-time flags to run `wasm-opt` over the output `wasm` file
* [ ] Enhance our test suite by running `wasm-opt` with most aggressive optimizations over everything, running both unoptimized/optimized versions, and comparing output.

The reason this issue comes to mind is that: `wasm-opt` provides a `--limit-segments` pass which merges the data segments. Our current segment layout algorithm doesn't do this, so for large programs (e.g. the `mini-hlint` example of `ghc-lib-parser`), the data segment count easily exceeds 100000 (the hard coded limit in V8). We can take a bit of time to do #289, but another option is running `wasm-opt` on it. Since `wasm` end-to-end optimization will be a thing in the future (similar to how people tries closure on ghcjs output), might as well start supporting it right now.

PR #390, opened 2019-12-12
Update `binaryen`

TerrorJack:
Changeset:

* Updates [`tweag/binaryen`](https://github.com/tweag/binaryen), rebasing upon upstream changes
* Updates the in-tree `binaryen` C++ sources to match `tweag/binaryen`
* Updates the raw bindings in `binaryen`
* Improves the custom `Setup.hs` of `binaryen`, removes the workaround for producing a self-contained `libbinaryen.a`, uses `absoluteComponentInstallDirs` to calculate the copy destination of the library component properly.
* Removes the `binaryen` Cabal flag of `asterius`, and get rid of all CPP macros
* Minor refactoring in the `binaryen` backend of `asterius`, dealing with API breaking changes

In preparation for #388


TerrorJack:
Would be nice to shave off the custom `Setup.hs` which calls `cmake` and replace it with Cabal's own C++ files support. Will give a try if the current version doesn't regress.

PR #399, opened 2019-12-16
Fix darwin support and test it on CI

TerrorJack:


TerrorJack:
There's an error (see https://circleci.com/gh/tweag/asterius/8576) when we attempt to use `runghc` to run `genapply`:

```
ghc-toolkit         > Configuring ghc-toolkit-0.0.1...
ghc-toolkit         > ghc: mmap 131072 bytes at 0x0: Operation not permitted
ghc-toolkit         > ghc: Try specifying an address with +RTS -xm<addr> -RTS
ghc-toolkit         > ghc: internal error: m32_allocator_init: Failed to map
ghc-toolkit         >     (GHC version 8.6.5 for x86_64_apple_darwin)
ghc-toolkit         >     Please report this as a GHC bug:  http://www.haskell.org/ghc/reportabug
ghc-toolkit         > readCreateProcess: /Users/distiller/.stack/programs/x86_64-osx/ghc-custom-asterius-8.6.5/bin/runghc "--ghc-arg=-Ighc-libdir/include" "--ghc-arg=-Iinclude-private" "genapply/Main.hs" (exit -6): failed
```

Of course, it's possible to remove the invocation of `genapply` and instead check in the generated `AutoApply.cmm` into the tree to simplify things. But I think that's some fundamental breakage we mustn't ignore.

ProofOfKeags:
How can I help with this ticket, I'd like to be able to use it on my Mac (the docker is a bit cumbersome), is the fix straight forward, would someone new to the codebase such as myself be able to move this forward?

TerrorJack:
@ProofOfKeags You can test building and booting `asterius` locally. Check the CircleCI [config](https://github.com/tweag/asterius/blob/wip-darwin-ci/.circleci/config.yml#L37) for example build steps.

We're currently blocked on an `ar`-related issue (`ar` in the darwin environment doesn't seem to support `rsp` file). We can workaround this by installing `binutils` from `brew` or by implementing a special `ar` wrapper program and telling `ahc-cabal` to use that. There may be other darwin-specific issues even after this one is solved. But you're welcome to give this a try and report back the errors you've encountered! And just fyi, the CircleCI darwin job is also triggered for forked PRs, so in case you'd like to fiddle with CI script instead of your local environment, that's also possible.

ProofOfKeags:
Yeah I tried to use the alternate ar from binutils but that causes regular stack builds for my other projects to fail, so some kind of switch seems necessary if that's the permanent route.

PR #403, opened 2019-12-17
Fix the request object name in default entry script to `req`

### LINKED ISSUE #364: Dashes in names of binaries make `ahc-dist` generate invalid javascript.
#### Description
The javascript generated by `ahc-dist` with the `--browser` flag is invalid, when the used executable has dashes in the name.
The resulting import will contain dashes as well, which are interpreted as subtraction operators.

#### Example output
```javascript
import test-project from "./test-project.req.mjs";
```

TerrorJack:
Closes #364.

We used to emit `import foo from "./foo.req.mjs"` in `foo.mjs` for `foo.hs` main modules, but when the module name contains slashes, `foo-bar` isn't a valid js identifier and this results in runtime syntax errors.

### DIFF
--- a/asterius/src/Asterius/Main.hs
+++ b/asterius/src/Asterius/Main.hs
@@ -217,18 +217,14 @@ genDefEntry task =
       "import module from \"./",
       out_base,
       ".wasm.mjs\";\n",
-      "import ",
-      out_base,
-      " from \"./",
+      "import req from \"./",
       out_base,
       ".req.mjs\";\n",
       case target task of
         Node -> "process.on(\"unhandledRejection\", err => { throw err; });\n"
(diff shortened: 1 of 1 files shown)


hsyl20:
Can we mangle the module name instead? Perhaps reusing z-encoding. It would allow several modules to live in the same directory (e.g. several diagrams examples).

TerrorJack:
Of course we can use Z-encoding here, but this shouldn't affect multiple main modules co-existing in the same directory. In all the generated files, `rts*.mjs` is always identical, and we only have `foo.req.mjs` and `foo.mjs` to contain module-specific code, and this PR only changes the `foo.mjs` content a little bit.

Issue #350, opened 2019-11-28
Properly split up `Asterius.Foreign`


TerrorJack:
`Asterius.Foreign` is the place where we paste GHC logic from `TcForeigns`/`DsForeigns` and hack them. The module is extremely large and hard to properly maintain, especially when we need to add/remove new hooked GHC functions, or migrating to a new GHC major version.

Things we need to do here:

* [x] Update our GHC fork, and export most things from `TcForeigns`/`DsForeigns`. Small export lists are good for upstream GHC, but they are a burden when we need to monkey patch stuff, since a lot of util functions need to be pasted
* [x] Split up `Asterius.Foreign` into `Asterius.TcForeign` and `Asterius.DsForeign`, and call upstream util functions whenever possible, removing all redundant util functions
* [ ] Add a doc section or comments describing the *what* & *why* of all our monkey patches, so to ease migrations to future GHC versions. A simple `diff` isn't sufficient.

TerrorJack:
Should have been closed by #406

Issue #294, opened 2019-09-15
Implement "foreign import wrapper" for JSFFI


TerrorJack:
Reborn from #49. This is necessary for exporting Haskell function closures at runtime, compared to `foreign export javascript`.

A side note: Anton Ekblad has a [paper](https://ekblad.cc/pubs/ifl15.pdf) describing a type class based mechanism of JSFFI for the haste compiler. Still unsure whether it's a superior approach compared to supporting the "foreign import wrapper" syntax, but it's an interesting read anyway, so I've recorded it here.

TerrorJack:
We also need to support this for the `ccall` convention, packages out there (e.g. `hslua`) require this to compile. Not opening a separate issue for the `ccall` part though.

PR #406, opened 2019-12-19
Workaround lack of "foreign import wrapper" support

TerrorJack:
Our `ghc` fork introduces a recent change: expose internal helper functions in FFI-related modules like `TcForeign`, `DsForeign` and `DsCCall`. In this way, we can cleaup the mess of pasted helper functions in `Asterius.Foreign` (see #350). This is a necessary requirement for "fixing" `foreign import wrapper` (at least for `ccall`) and unblocking packages like `hslua` and `pandoc`.

Work in progress since `ghc` needs to change in lockstep, don't merge yet.


TerrorJack:
This PR now also workarounds our previous lack of `foreign import "wrapper"` support. Since this FFI construct is used to marshal Haskell functions to `FunPtr`s, but we haven't figured out proper semantics of `FunPtr` and its relation to the JS world yet, so right now, the implementation simply inserts a runtime error as the generated RHS for such imports.

This should be sufficient to unblock packages like `hslua`. A more proper fix for #294 takes more work and is scheduled for later.

Issue #290, opened 2019-09-15
Fix and test the Cloudflare Worker example


TerrorJack:
There's a `cloudflare` unit test that emits code for a Haskell module that's supposed to be run on Cloudflare Worker. It was a very simple POC to demonstrate Cloudflare Worker support last year, manually tested with my own Cloudflare subscription, but not on CI. A lot of breaking changes in the project happened since then, most notably, RTS API are all made async and the "sync mode" is removed, thus breaking the existing `cloudflare` example.

Now we have a much more decent runtime, it would be a good timing to fix the cloudflare example and test it on CI. There's a [project](https://github.com/dollarshaveclub/cloudworker) to mock the Worker runtime.

Note about "sync mode": Cloudflare Worker runtime requires that when the JavaScript file is run, the request handler must be registered *synchronously*, so in the past we did a "sync mode". But since the handler itself can be async, we can still supply a dummy handler on the first run, and the real initialization fulfills a global `Promise` used by the handler.

It may also be beneficial to add a Cloudflare Worker backend in the linker, along with the Node/Browser backends, to automate the creation and bundling of special entry scripts.

TerrorJack:
Closed by #413.

PR #413, opened 2020-01-06
Fix the Cloudflare Worker example and enable its test in CI

ento:
This PR attempts to address most of #290.

There is a bit more JS code in `cloudflare.mjs` to hande the initialization of Asterius's runtime and the `fetch` event listener.

`cloudflare.hs` exports a handler function instead of having a `main` function. (Note: the handler looks identical to [this template project's](https://github.com/ento/worker-haskell-template/blob/3b393e2027013fbbe2b749047508a912de5c1108/src/worker.hs) because both this PR and that template project grew organically out of my attempt to use Asterius to write a Cloudflare Workers script.)

There is also a fair bit of JS code in `test.js` to test the compiled code using "@dollarshaveclub/cloudworker". This _could_ be written in Haskell, which would launch `cloudworker` as a child process and send test requests against it. The choice of using JS here is due to my inexperience with that kind of testing in Haskell.

I did not add `Cloudflare` to the `Asterius.Main.Task.Target` type in this PR, because it worked without doing so. The benefits I can see right now are:

1. The step to generate HTML can be skipped
2. The step to generate `.wasm.mjs` can be skipped, if we keep the current setup of letting the user write the entire entry script.
   - Alternatively, `.wasm.mjs` can export a Promise that resolves to a new Asterius instance (`asteriusInstance` in this PR's code). In this case, the compiler will need to know -- probably through a command line flag -- the global variable name (binding name) under which Cloudflare Workers runtime will make the Wasm module available. As noted in `cloudflare.mjs` below, the binding name is configurable by the user when using the API. It gets hardcoded to 'wasm' when using the Wrangler CLI (as of v1.6.0).

### DIFF
--- a/asterius/test/cloudflare.hs
+++ b/asterius/test/cloudflare.hs
@@ -14,8 +14,7 @@ main = do
       "--input-mjs",
       "test/cloudflare/cloudflare.mjs",
       "--export-function=handleFetch",
-      "--ghc-option=-no-hs-main",
-      "--extra-root-symbol=Worker_x_closure"
+      "--ghc-option=-no-hs-main"
     ]
       <> args
   withCurrentDirectory "test/cloudflare" $ callProcess "npm" [ "run", "test" ]
--- a/asterius/test/cloudflare.hs
+++ b/asterius/test/cloudflare.hs
@@ -11,6 +11,9 @@ main = do
       "--input-hs",
       "test/cloudflare/cloudflare.hs",
       "--input-mjs",
-      "test/cloudflare/cloudflare.mjs"
+      "test/cloudflare/cloudflare.mjs",
+      "--export-function=handleFetch",
+      "--ghc-option=-no-hs-main",
+      "--extra-root-symbol=Worker_x_closure"
     ]
       <> args
--- a/asterius/test/cloudflare.hs
+++ b/asterius/test/cloudflare.hs
@@ -5,6 +5,7 @@ import System.Process
 main :: IO ()
 main = do
   args <- getArgs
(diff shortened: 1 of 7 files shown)


ento:
I'm seeing this error from a published script on every other request. I've pushed an update to the test suite that reproduces the error for now; will look into it later.

> TypeError: Body has already been used. It can only be used once. Use tee() first if you need to read it twice.

ento:
517763c should've fixed the "Body has already been used" issue above. Waiting for CI...

TerrorJack:
@ento Thanks for the work! Is this PR ready for review?

ento:
@TerrorJack Yes, it is ready for review.

asterius/test/cloudflare.hs
```diff
@@ -5,12 +5,17 @@ import System.Process
...
     [ "--bundle",
       "--browser",
       "--input-hs",
       "test/cloudflare/cloudflare.hs",
       "--input-mjs",
-      "test/cloudflare/cloudflare.mjs"
+      "test/cloudflare/cloudflare.mjs",
+      "--export-function=handleFetch",
+      "--ghc-option=-no-hs-main",
+      "--extra-root-symbol=Worker_x_closure"
```
TerrorJack: There doesn't seem to be a `Worker` module in this test, nor is there a top-level `x` closure?

ento: This flag was added ritualistically when trying to make `-no-hs-main` work, without understanding of what it actually does. I've removed the line and confirmed that tests pass.

ento:
Rebased on master and force-pushed.

Issue #407, opened 2019-12-19
Factor out the platform-specific parts of the runtime


TerrorJack:
Closely related: #402 

The runtime should be restructured as follows:

* The existing `rts.mjs` and all `rts.*.mjs` modules are all platform-independent, and work on both node and browser. The platform-specific interfaces are referenced through an extra "capability" object passed to the constructors whenever needed.
* The `rts/node` and `rts/browser` directories contain platform-specific logic. `rts/platform/default.mjs` should contain a default export which is the "capability" object mentioned above.
* `rts.mjs` stays self-contained. The default entry script would either import from `rts/node` or `rts/browser` and use the default capability accordingly.
* The TH runner and `--bundle` logic needs to be adjusted accordingly.

The refactorings proposed above do not require introducing CPP into the runtime for platform-specific functionality, preserve the self-contained property of `rts.mjs` so it can still be reused across different asterius instances, and do not introduce additional config files and packagers like `webpack`. The downside is slightly affected performance (compared to ghcjs's approach of CPPs and global `h$` functions) due to indirect calls, but the cost should barely be noticable.

TerrorJack:
Closed by #414.

PR #414, opened 2020-01-06
Implementing Performance API

acondolu:
This is work in progress to enable the asterius runtime to import a different module according to the deployment target (for now, either Node.js or the browser).

This target-specific module should provide a target-independent interface for methods useful to other classes: for instance, the Performance API that we plan to use for recording statistics about garbage collection, which is currently available in browsers and node.js in a different way.

The current commit adds two new folders to the rts, `browser` and `node`, each containing a target-specific `default.mjs` file (currently a stub, to be filled later). The `target` attribute in a `Task` object determines which folder to copy at link time. 

The target-specific import is obtained by a dynamic import performed inside the JS `req` object, and stored in the `targetSpecificModule` attribute.

### DIFF
--- a/asterius/rts/browser/default.mjs
+++ b/asterius/rts/browser/default.mjs
@@ -9,7 +9,7 @@ export default {
     /**
      * Returns the current timestamp, where 0 represents
      * the time origin of the document.
-     * The output is a [seconds, nanoseconds] Array.
+     * @returns A [seconds, nanoseconds] Array.
      */
     getCPUTime: () => {
       const ms = performance.now(),
@@ -19,7 +19,7 @@ export default {
     },
     /**
      * Returns the current timestamp, where 0 represents UNIX Epoch.
-     * The output is a [seconds, nanoseconds] Array.
+     * @returns A [seconds, nanoseconds] Array.
      */
     getUnixEpochTime: () => {
       const ms = Date.now(),
--- a/asterius/rts/browser/default.mjs
+++ b/asterius/rts/browser/default.mjs
@@ -5,27 +5,30 @@ export default {
   /**
    * The Performance Web API.
    */
-  Performance: window.performance,
+  Performance: performance,
   /**
    * A custom Time interface, used in {@link TimeCBits}.
    */
   Time: {
     /**
-     * Returns the current millisecond timestamp,
-     * where 0 represents the time origin of the document.
+     * Returns the current timestamp, where 0 represents
+     * the time origin of the document.
+     * The output is a [seconds, nanoseconds] Array.
      */
-    now: () => {
-      return performance.now();
+    getCPUTime: () => {
+      const ms = performance.now(),
+            s = Math.floor(ms / 1000.0),
+            ns = Math.floor(ms - s * 1000) * 1000000;
+      return [s, ns];
     },
     /**
-     * Returns the current millisecond timestamp,
-     * where 0 represents UNIX Epoch. The output is a
-     * [seconds, nanoseconds] Array.
+     * Returns the current timestamp, where 0 represents UNIX Epoch.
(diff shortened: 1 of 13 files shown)


TerrorJack:
This issue should close #407

acondolu:
The last commit replaces the dynamic with a static import, and reverts the `request` object to a usual dictionary (i.e. not a `Promise` anymore). 

Moreover, it extends the (previously stub) interface of the target-specific module with time and Performance APIs. This PR improves on #369 by exposing high-resolution timestamps when available (i.e., on node.js) and using them in the `TimeCBits` class.

asterius/rts/browser/default.mjs
```diff
@@ -0,0 +1,41 @@
+/**
+ * This module contains browser-specific functionality.
+ */
+export default {
+  /**
+   * The Performance Web API.
+   */
+  Performance: window.performance,
```
TerrorJack: The `Performance` field exported by both `browser` & `node` default script doesn't seem to be used anywhere. Also, I suggest `performance` in this script instead of explicit `window.performance`, since `performance` can also be a field of `WorkerGlobalScope`, in case the script is run inside a web worker.

asterius/rts/rts.time.mjs
```diff
@@ -1,32 +1,52 @@
...
-        s = Math.floor(ms / 1000),
-        ns = (ms - s * 1000) * 1000000;
-      this.memory.i64Store(tp + rtsConstants.offset_timespec_tv_sec, s);
-      this.memory.i64Store(tp + rtsConstants.offset_timespec_tv_nsec, ns);
+  /**
+   * Stores at a memory address the resolution of a given clock.
+   * @param clk_id the id of the clock
+   * @param addr the memory address 
+   */
+  clock_getres(clk_id, addr) {
```
TerrorJack: Now that we have `now` for monotonic clock and `time` for UTC clock, it would be nice to start using `clk_id` for both `clock_getres` and `clock_gettime` to know which clock we're querying, and use results from different underlying clocks. We need to know the constants for `CLOCK_MONOTONIC` and `CLOCK_REALTIME` here, and the rough steps to use those constants are:

1. Adding relevant constants in `cbits` of `ghc-toolkit`, and their foreign imports
2. Modify the `Asterius.JSGen.Constants` to emit those constants into `rts.constants.mjs` script, and run `genconstants` to generate new script to update the one in the project git tree
3. Import and use `rts.constants.mjs` file here

asterius/rts/rts.time.mjs
```diff
@@ -1,32 +1,52 @@
...
+    if (addr) {
+      var sec = 0, nsec = this.resolution;
```
TerrorJack: `let` & `const` are better ways of introducing bindings in JS these days.

asterius/src/Asterius/Main.hs
```diff
@@ -406,9 +416,13 @@ ahcDistMain logger task (final_m, report) = do
     "[INFO] Writing JavaScript runtime modules to "
       <> show
         (outputDirectory task)
-  rts_files <- listDirectory $ dataDir </> "rts"
+  rts_files' <- listDirectory $ dataDir </> "rts"
+  let rts_files = filter (\x -> x /= "browser" && x /= "node") rts_files'
   for_ rts_files $
     \f -> copyFile (dataDir </> "rts" </> f) (outputDirectory task </> f)
+  let specificFolder = map toLower $ show $ target task
```
TerrorJack: The function mapping from a `Target` to the directory name would be nice to be factored out here. Indeed `map toLower . show` works here but good to avoid duplication a little bit, and IMO a plain pattern matching is better.

acondolu: OK, rolled back to pattern matching

PR #417, opened 2020-01-06
Add cmath cbits

TerrorJack:
This PR adds implementation of the `scalbn` function in libc, which is used by `__word_encodeDouble` in `integer-simple`.

The code in `CMath.hs` doesn't need to be reviewed; it is expression IR without using EDSL, hand-written by peeking the disassembled output of `wasi-sdk` compiled `musl`. We really need more principled automatic disassembling to make use of more libc functions in the long run though; I definitely don't want to do this twice

### DIFF
--- a/asterius/src/Asterius/Builtins/CMath.hs
+++ b/asterius/src/Asterius/Builtins/CMath.hs
@@ -10,9 +10,34 @@ where
 
 import Asterius.Types
 
+-- | Here contains implementation of certain functions in math.h of libc. They
+-- are written in the expression IR directly without using our EDSL, and are not
+-- meant to be understood by humans. For the curious souls, here's the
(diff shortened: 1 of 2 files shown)


asterius/src/Asterius/Builtins/CMath.hs
```diff
@@ -0,0 +1,247 @@
...
+  ( cmathCBits,
+  )
+where
+
+import Asterius.Types
+
+cmathCBits :: AsteriusModule
+cmathCBits = scalbn
+
+scalbn :: AsteriusModule
```
hsyl20: Could you add a comment here indicating the provenance of this code (as in the MR description) and why we shouldn't try to understand it?

PR #421, opened 2020-01-07
No-op threadPaused

TerrorJack:
Undoes #145 and #149 

In the GHC rts, there's a `threadPaused` function (see `rts/ThreadPaused.c`) which is called when a Haskell thread is about to suspend due to calling a `safe` import; it walks the STG stack and performs the "lazy blackholing" optimization described in the "Haskell on a Shared-Memory Multiprocessor" paper, claiming thunk objects as blackholes, and strips stack segments related to existing blackholes (which indicates thunks claimed by other threads) to reduce duplicated computation.

We implemented `threadPaused` stack walking and lazy blackholing logic before. However, removing this logic isn't supposed to break anything, since we don't have a SMP runtime anyway, so the race condition of two threads evaluating the same thunk isn't a concern. It's still possible that a Haskell thread is suspended due to GC, and an update frame is on that thread's STG stack, pointing to a thunk which is then forced by another thread, but that's fine.


hsyl20:
Does it improve the runtime performance?

TerrorJack:
> Does it improve the runtime performance?

Very slightly yes, since we avoid walking the stack during a safe ccall, and the previous behavior of lazy blackholing doesn't improve performance for non-SMP runtimes. But the point is mainly pulling some weed from the runtime.

PR #423, opened 2020-01-10
(WIP) More comments in the RTS source code

acondolu:
This PR adds more comments to the source code of the asterius RTS, in particular to the source files related to memory management and garbage collection (i.e. `rts.memory.mjs`, `rts.heapalloc.mjs`, `rts.gc.mjs`).
The goal is to clarify the current logic of GC, in order to make more straightforward hacking garbage collection in the future so to improve performance.

The added comments are in JSDoc format, so that documentation can later be easily generated.

### DIFF
--- a/asterius/rts/rts.gc.mjs
+++ b/asterius/rts/rts.gc.mjs
@@ -43,26 +43,38 @@ export class GC {
     this.yolo = yolo;
     /**
      * Garbage collection will not be performed when the
-     * current number of MBlocks is less than {@link GC#gcThreshold}
-     * (see {link GC#performGC}).
+     * current number of "live" MBlocks is less than
+     * {@link GC#gcThreshold} (see {@link GC#performGC}).
      * @name GC#gcThreshold
      * @default 64
      */
     this.gcThreshold = gcThreshold;
     /**
-     * Map used during evacation, in order to store
+     * Map used during evacuation, in order to store
      * the forwarding pointers from original objects to their copies
-     * (see {@link GC#evacuateClosure}).
+     * (see {@link GC#evacuateClosure}). Note: the pointers are 
+     * stored without their dynamic pointer tag (i.e. they have 
+     * been {@link Memory#unDynTag})-ed beforehand).
      * @name GC#closureIndirects
      */
     this.closureIndirects = new Map();
     /**
-     * Set containing the objects in the to-space
+     * Set containing the MBlocks in the to-space,
+     * i.e. the MBlocks where reachable objects are copied
+     * during garbage collection.
+     * Notes:
+     * 1) Pinned MBlocks are not copied during GC: they are
+     *    simply set as live, and added to the liveMBlocks set.
+     * 2) Static objects are not copied either, but their
+     *    blocks are not even added to the liveMBlocks set.
      * @name GC#liveMBlocks
      */
     this.liveMBlocks = new Set();
     /**
-     * Set containing the object in the from-space
+     * Set containing the MBlocks in the from-space,
+     * i.e. the MBlocks that have containing objects
+     * that have been copied into to-space. These MBlocks
+     * will be freed at the end of garbage collection.
      * @name GC#deadMBlocks
      */
     this.deadMBlocks = new Set();
@@ -96,7 +108,7 @@ export class GC {
   }
 
   /**
-   * Heap alloactes a phisical copy of the given closure.
+   * Heap alloactes a physical copy of the given closure.
    * Used during evacuation by {@link GC#evacuateClosure}.
    * @param c The source address of the closure
    * @param bytes The size in bytes of the closure 
@@ -112,8 +124,8 @@ export class GC {
   /**
    * Evacuates a closure. This consists of: 
    * (1) Copying the closure into to-space through {@link GC#copyClosure}
-   * (2) Map the old address of the closure to its new address
-   *     in {@link GC#closureIndirects}.
+   * (2) Map the old unDynTag-ed address of the closure
+   *     to its new unDynTag-ed address in {@link GC#closureIndirects}.
    * If that closure had already been evacuated, simply
    * return the forwarding pointer already present in {@link GC#closureIndirects}.
    * @param c The memory address of the closure to evacuate.
@@ -316,7 +328,7 @@ export class GC {
       } else {
         // If the closure belongs to the static part of the memory,
         // we do not actually copy it into to-space, but we still set
-        // it to evacuated in order to avoid future checks.
+        // it to evacuated and we enqueue it for scavenging.
         dest_c = untagged_c;
       }
       // Add a forwarding pointer from the original closure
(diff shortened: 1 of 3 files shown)


asterius/rts/rts.gc.mjs
```diff
@@ -32,22 +35,72 @@ export class GC {
...
+    /**
+     * 'Yolo' mode disables garbage collection altogether
+     * (see {@link GC#performGC})
+     * @name GC#yolo
+     */
     this.yolo = yolo;
-    this.gcThreshold = gc_threshold;
+    /**
+     * Garbage collection will not be performed when the
+     * current number of MBlocks is less than {@link GC#gcThreshold}
```
TerrorJack: Or more precisely, when the current number of "live" MBlocks (as returned by `this.heapAlloc.liveSize()`). "live" means they were allocated by `getMBlocks()` some time ago and hasn't been `freeMBlocks()` yet. Also, this duplicates a bit with the docs at the beginning of `performGC()`, so maybe we should point to that place and improve the description there?

asterius/rts/rts.gc.mjs
```diff
@@ -32,22 +35,72 @@ export class GC {
...
+     * (see {link GC#performGC}).
+     * @name GC#gcThreshold
+     * @default 64
+     */
+    this.gcThreshold = gcThreshold;
+    /**
+     * Map used during evacation, in order to store
```
TerrorJack: Worth mentioning that the from/to pointers are both `unDynTag`ed (removed lowered 3 bits used by GHC's dynamic pointer tagging).

asterius/rts/rts.gc.mjs
```diff
@@ -32,22 +35,72 @@ export class GC {
...
+     * @name GC#closureIndirects
+     */
     this.closureIndirects = new Map();
+    /**
+     * Set containing the objects in the to-space
+     * @name GC#liveMBlocks
+     */
     this.liveMBlocks = new Set();
+    /**
+     * Set containing the object in the from-space
```
TerrorJack: There are pinned and unpinned objects in the fromspace. Only the unpinned ones gets moved and have their original mblocks marked as dead, and the pinned ones' original mblocks are still live. Worth mentioning here.

asterius/rts/rts.gc.mjs
```diff
@@ -32,22 +35,72 @@ export class GC {
...
+   * @param addr The memory address to check
+   */
+  isPinned(addr) {
+    const bd = bdescr(addr),
       flags = this.memory.i16Load(bd + rtsConstants.offset_bdescr_flags);
     return Boolean(flags & rtsConstants.BF_PINNED);
   }
 
+  /**
+   * Heap alloactes a phisical copy of the given closure.
```
TerrorJack: /s/phisical/physical

asterius/rts/rts.gc.mjs
```diff
@@ -56,22 +109,37 @@ export class GC {
     return dest_c;
   }
 
+  /**
+   * Evacuates a closure. This consists of: 
+   * (1) Copying the closure into to-space through {@link GC#copyClosure}
+   * (2) Map the old address of the closure to its new address
```
TerrorJack: The old/new `unDynTag`ed address.

asterius/rts/rts.gc.mjs
```diff
@@ -244,9 +314,16 @@ export class GC {
           }
         }
       } else {
+        // If the closure belongs to the static part of the memory,
```
TerrorJack: Indeed we avoided future checks, but imo the real point here is to enable the code below to push the static closure to the scavenging work list. We don't evacuate static objects, but we do scavenge them.

asterius/rts/rts.gc.mjs
```diff
@@ -657,38 +743,54 @@ export class GC {
     }
   }
 
+  /**
+   * Allocates a new nursery and stores its address in the appropriate
+   * field of the StgRegTable of the main capability.
+   */
   updateNursery() {
+    // Note: the 'rHpAlloc' field of the 'StgRegTable' C struct contains 
+    // the number of bytes being allocated in the heap. Here, this field is
```
TerrorJack: More precisely, `HpAlloc` contains the number of bytes being attempted to allocate before the heap check fails.

asterius/rts/rts.gc.mjs
```diff
@@ -657,38 +743,54 @@ export class GC {
...
+    // read in the hp_alloc variable and used as the size of the newly
```
TerrorJack: "As the size" isn't quite right. E.g. if we're allocating a closure of 8 bytes, and we ran out of nursery space, but we can't have a nursery with exactly 8 bytes' size. So "used to determine the size" shall be more precise.

asterius/rts/rts.gc.mjs
```diff
@@ -657,38 +743,54 @@ export class GC {
...
         this.symbolTable.MainCapability + rtsConstants.offset_Capability_r,
       hp_alloc = Number(
         this.memory.i64Load(base_reg + rtsConstants.offset_StgRegTable_rHpAlloc)
       );
+    // reset the number of allocated bytes in the nursery
     this.memory.i64Store(
       base_reg + rtsConstants.offset_StgRegTable_rHpAlloc,
       0
     );
+    // The address of the new nursery is stored in the 'rCurrentNursery'
```
TerrorJack: The address of the new nursery's block descriptor.

asterius/rts/rts.heapalloc.mjs
```diff
@@ -1,13 +1,42 @@
 import * as rtsConstants from "./rts.constants.mjs";
 
+/**
+ * Class implementing the allocation of heap objects.
```
TerrorJack: Implementing the allocation of nurseries, and also individual heap objects.

asterius/rts/rts.heapalloc.mjs
```diff
@@ -1,13 +1,42 @@
...
+ * This mainly consists of a block allocator, which allocates
```
TerrorJack: In the current asterius rts, we do not really distinct between mblocks and blocks, and thus all blocks are really mblocks.

asterius/rts/rts.heapalloc.mjs
```diff
@@ -1,13 +1,42 @@
...
 export class HeapAlloc {
   constructor(memory) {
+    /**
+     * @type Memory
+     * @name HeapAlloc#memory
+     */
     this.memory = memory;
+    /**
+     * An array with two entries:
+     * 1. The unpinned pool, i.e. the address of a MBlock
```
TerrorJack: The block descriptor address.

asterius/rts/rts.memory.mjs
```diff
@@ -4,17 +4,58 @@ function mask(n) {
...
     this.staticMBlocks = undefined;
+    /**
+     * Low-level interfaces for reading/writing the contents
+     * of {@link Memory#memory}.
+     * @name Memory#i8view
+     * @name Memory#dataView
+     */
     this.i8View = undefined;
     this.dataView = undefined;
+    /**
```
TerrorJack: Maybe we should explain further what "capacity" means here. We separate between "size" and "capacity", and when we need to grow the underlying wasm linear memory, we always grow it at least by a factor of 2, to reduce amortized overhead of allocating individual mblocks; this is a well known technique of implementing things like `std::vector` in C++. So "capacity" is the real size of wasm linear memory, and size indicates the range within the capacity that we really use at the moment.

asterius/rts/rts.gc.mjs
acondolu: Added the precisation. This one comment is still a bit duplicated, but I guess _repetita iuvant_ 😅

PR #424, opened 2020-01-13
Use forwarding pointers during GC

acondolu:
During copying garbage collection, the GHC runtime employs so-called "forwarding pointers", i.e. it replaces the header word of the object that is being copied with a pointer to the new copy. In addition to overwriting the header, the least-significant bit must be set to 1, in order to be able to distinguish a forwarding pointer from a pointer to the info table.

The `asterius` RTS currently does not exploit forwarding pointers, and instead uses `GC.closureIndirects` to map the address of an evacuated object to its copy.

This PR adds support for forwarding pointers to `asterius`.

### DIFF
--- a/asterius/rts/rts.gc.mjs
+++ b/asterius/rts/rts.gc.mjs
@@ -171,6 +171,7 @@ export class GC {
         throw new WebAssembly.RuntimeError(
           `Invalid info table 0x${info.toString(16)}`
         );
+      // Get the type of the closure from info tables
       const type = this.memory.i32Load(
         info + rtsConstants.offset_StgInfoTable_type
       );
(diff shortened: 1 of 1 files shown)


TerrorJack:
Note that we have `155` failures on the `asterius-test-ghc-testsuite` job, increasing from `149` on `master`. Needs further investigation before merging.

acondolu:
I should have fixed the regression in the last commits.

I could not get rid completely of the `GC.closureIndirects` map: I still have a `GC.nonMovedObjects` list that keeps track of the objects that are not copied during evacuation (pinned objects and static objects). During evacuation, the header of these objects is still overwritten with a forwarding pointer (to its address itself), but this actually overwrites the info pointer: some extra care is taken to pass the original info pointer to the scavenging function, and to handle it correctly.

PR #433, opened 2020-01-23
Tuning ghc testsuite timeout

TerrorJack:
This PR adjusts the `tasty` timeout for individual test cases of `ghc-testsuite`. Currently, when we try to diff the CSV reports of `debug` and `vanilla` flavours of the test run, the additional failures in the `debug` report are all simply `AsyncCancelled`, which is most likely caused by `tasty` timeouts.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -261,7 +261,7 @@ jobs:
 
 
               # run test cases that can fail.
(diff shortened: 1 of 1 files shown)


TerrorJack:
Even after bumping individual test timeout from `180s` to `600s`, the additional failures in the `debug` flavour is still `AsyncCancelled`, which suggests they are either infinite loops at the wasm end, or there's some deeper problems there. So tuning the timeouts isn't helpful in exposing new bugs, closing this one.

PR #436, opened 2020-01-27
Tune rtsopts for asterius executables

TerrorJack:
This PR tunes the rtsopts in the CI script and `asterius` executables:

* We used to use `GHCRTS` environment variable to pass `-N` to the `asterius` executables. This optimization does more harm than good:
  * We had to patch `ghc-pkg` and `hsc2hs` since their executables aren't linked with `-threaded -rtsopts`
  * `ahc-cabal` already has package-level parallelism when booting
* `GHCRTS` is removed in the jobs other than `ghc-testsuite`. This also slightly reduce our work to maintain the forked ghc, since we don't need to work with an `hsc2hs` fork anymore
* The `asterius` executables are still linked with the threaded rts, and additionally with larger default nursery size, with parallel gc and idle gc disabled.

The combined effect is slightly improved CI and docker image build time. And it also helps with our procedure of upgrading to new ghc major versions.

### DIFF
--- a/asterius/package.yaml
+++ b/asterius/package.yaml
@@ -49,6 +49,9 @@ custom-setup:
 
 ghc-options: -Wall
 
+_exe-ghc-options: &exe-ghc-options
+  - -threaded -rtsopts "-with-rtsopts=-A64m -n2m -I0 -qg"
+
 dependencies:
   - base
   - binary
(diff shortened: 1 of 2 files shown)


asterius/package.yaml
```diff
@@ -100,49 +100,49 @@ executables:
   ahc:
     source-dirs: app
     main: ahc.hs
-    ghc-options: -threaded -rtsopts
+    ghc-options: -threaded -rtsopts "-with-rtsopts=-A64m -n2m -I0 -qg"
```
hsyl20: Couldn't we use a top-level `ghc-options` used by all the executables instead of copy-pasting it?

TerrorJack: iirc that will emit GHC warnings for the library target.

hsyl20: We could use anchors and aliases as mentioned in the docs: https://github.com/sol/hpack#not-repeating-yourself The example they give is exactly what we want to do.

PR #439, opened 2020-01-27
Implement selector optimisation

### LINKED ISSUE #422: Fix the selector thunk optimization
See https://gitlab.haskell.org/ghc/ghc/wikis/commentary/rts/storage/heap-objects#selector-thunks for a description of "selector thunk" in ghc rts.

The current garbage collector treats `THUNK_SELECTOR` objects just like ordinary objects, simply copying it to the to-space and evacuating its selectee. As described in the link above, this can be a minor source of space leaks, and we should implement correct selector thunk optimization in the gc.

acondolu:
_Selector optimisation_ is a known optimisation where thunks of applications of field selectors are evaluated during garbage collection, so to avoid space leaks caused by unused fields that look alive.

This PR fixes #422 by introducing the new function `GC.stingyEval`, which shorts-out chains made out of selector/constructor and/or indirections (also interleaved).

Note: `GC.stingyEval` uses a technique called _whiteholing_, where closures in the chain that is being explored are replaced with "whiteholes", in order to not diverge in case of cyclic closures.

- [x] Expose the necessary info pointers as `AsteriusEntitySymbol`s
- [x] Implement whiteholing
- [x] Double-check propagation of dynamic tags

### DIFF
--- a/asterius/rts/rts.gc.mjs
+++ b/asterius/rts/rts.gc.mjs
@@ -120,6 +120,117 @@ export class GC {
     return dest_c;
   }
 
+/**
+ * Performs _stingy_ evaluation, i.e. a very frugual form
+ * of evaluation that is carried during garbage collection.
+ * It implements the following two optimizations:
+ * - Indirections short-cutting;
+ * - Selector optimization: remove thunks of applications of field
+ *   selectors.
+ * Only the argument `c` is required: the other arguments will be
+ * computed when `undefined`.
+ * @param c - The address of the closure
+ * @param {number=} untagged_c - The unDynTag-ed address
+ * @param {number=} info - The info pointer of `c`
+ * @param {number=} type - The closure type of `c`
+ * @returns A tuple array `[header, type]` where `header` is the header
+ *   of the closure `c` after evaluation, and `type` is the resulting
+ *   closure type.
+ */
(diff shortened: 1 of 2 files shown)


asterius/rts/rts.gc.mjs
```diff
@@ -120,6 +120,117 @@ export class GC {
...
+          }
+          case ClosureTypes.THUNK_SELECTOR: {
+            if (res_type == ClosureTypes.CONSTR) {
+              const offset = this.memory.i32Load(
+                res_info + rtsConstants.offset_StgInfoTable_layout
+              );
+              peek[2] = this.symbolTable.stg_IND_info;
+              stack.push(peek); 
+              // Start evaluating the selected field
+              stack.push([res_c + offset,,,]);
```
curiousleo:
```javascript
              stack.push([res_c + offset,,,]);
```

What's this `,,,` syntax?

acondolu: `[res_c + offset,,,]` means an array with four entries, the last three are `undefined`. I made explicit the `undefined` entries just because only arrays that are quadruples are pushed to the `stack`.

curiousleo: Ah, thanks for the explanation!

PR #440, opened 2020-01-30
Update rts.unicode.mjs

TerrorJack:
This PR uses the util script in #335 to update the in-tree `rts.unicode.mjs` artifact, to catch up with relevant changes in `ghc-8.8.2`.


curiousleo:
:grinning:

PR #454, opened 2020-02-13
Recording mutated closures

acondolu:
#### Context
This PR is a prerequisite for implementing generational garbage collection. In fact, generational GC requires to keep track of all the closures in the older generation that contain a pointer to a closure in the younger generation. This means that whenever a closure in the older generation is mutated, we need to record it in a set which is usually called `remembered set` or `mut_list`.

#### Changes
The types of the closures that can mutated at runtime are: Thread State Objects (`TSO`), `STACK`, synchronising variables (`MVAR`), mutable variables (`MUT_VAR`), arrays with pointers (`MUT_ARR_PTRS`, `SMALL_MUT_ARR_PTRS`). For each of these closure types, we need to ensure that the `PrimOps` functions manipulating them record correctly their mutation. We provide a stub `GC.recordMutated` method, which will enqueue mutated closures in a remembered set (a future PR).

#### Work in progress
Do not merge yet.


acondolu:
Before merging, we should ensure that `updateThunk` (and probably also `messageBlackHole`, but I'm not very familiar with that that function does) call `recordMutable` when blackholing. The problem is that `Blackhole.hs` is imported in `Builtins.hs`, so we need some refactoring before, to avoid the cyclic dependency.

PR #459, opened 2020-02-24
Fix Selector Optimisation, pt. II

acondolu:
The PR #439 implemented selector optimisation during GC, but it also included a few bugs that went undetected (mainly because the optimisation is never really performed by the tests in `ghc-testsuite`, and also because we were not able to create artificial unit tests that covered the whole code).

The cause for the errors was basically in a few pointers that were not dereferenced correctly, an issue in the selection itself, and the incomplete handling of all constructor closure types (`CONSTR`, `CONSTR_i_j`, `CONSTR_NOCAF`).

This PR also provides more comments, and a refactoring that uses locally defined variables, so to avoid a previous bug due to variable shadowing.

**Note.** The `stingyEval` function, which performs the optimisation, is a non-trivial function that keeps a stack of encountered closures: in this way we avoid the function to be recursive, but the code becomes less easy to understand.

### DIFF
--- a/asterius/rts/rts.gc.mjs
+++ b/asterius/rts/rts.gc.mjs
@@ -129,161 +129,101 @@ export class GC {
    *   selectors.
    * Only the argument `c` is required: the other arguments will be
    * computed in case they are `undefined`.
-   * @param c - The address of the closure
+   * @param {number} c - The address of the closure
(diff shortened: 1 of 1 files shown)


acondolu:
👍 I have just pushed a recursive version of `stingyEval`. It is quite more short and easy to understand than the previous one!

PR #468, opened 2020-02-28
Fix static pointers support

TerrorJack:
This PR fixes static pointers.

* In the `ghc-toolkit` IR, we add `SptEntry` list so the codegen can process them
* In the `asterius` IR, we add a mapping from static closure symbols to the SPT fingerprints. The mapping is added when we process raw Cmm compiled from Haskell
* In the linker, after the regular gc-sections pass, we pick out the subset of all SPT entries where the static closures are alive, and emit a mapping as `req.sptEntries` in the req script
* The RTS maintains a mapping from 128-bit `Fingerprint`s to static closure addresses. Upon startup, each SPT entry also gets a `StablePtr` allocated, so they become GC roots
* We needed to slightly patch `GHC.StaticPtr` in order to implement `staticPtrKeys`, since the fingerprint contents needs to be passed to a buffer allocated in the Haskell land

### DIFF
--- a/asterius/rts/rts.mjs
+++ b/asterius/rts/rts.mjs
@@ -11,6 +11,7 @@ import { MemoryTrap } from "./rts.memorytrap.mjs";
 import { HeapAlloc } from "./rts.heapalloc.mjs";
 import { StablePtrManager } from "./rts.stableptr.mjs";
 import { StableNameManager } from "./rts.stablename.mjs";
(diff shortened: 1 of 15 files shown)


TerrorJack:
When `wasm-toolkit` also supports custom sections in the future, we shall seek to emit `sptEntries` as a custom section, to further reduce overall code size.

PR #486, opened 2020-03-12
Fix ghcid support

TerrorJack:
This PR fixes `ghcid` support for the `asterius:lib` target. If we attempt to load the target via `stack ghci` then somehow the `ghci` linker chokes upon unresolved symbols when trying to load `binaryen`; now we workaround the problem by using `stack exec ghci` and manually loading the source files.

How to use:

* Do a `stack build asterius` first. No need for booting.
* Run `utils/ghcid.sh` from the project root directory. That's it.

The experience is so much better than `stack build --file-watch`.

In this PR we also removed the custom `Setup.hs` script of the `asterius` package. A minor side-effect: we don't hard-code the host `ghc` and `ghc-pkg` paths anymore, but they should be correctly resolved at runtime if things are run with `stack exec` or `stack test`, or when `direnv` is in effect.


TerrorJack:
Another fix that sneaked in but I didn't bother to make a separate PR: we now pin specific `apt` repo snapshots for improved reproducibility of builds.

Issue #522, opened 2020-03-30
Use a Reader monad in Asterius.Backends.WasmToolkit


gkaracha:
Similar situation as in Issue https://github.com/tweag/asterius/issues/515: Several read-only arguments are passed to `makeInstructions`, making it (a) longer and (b) more difficult to read. Also, extending this list of parameters means that we have to touch a lot of code unnecessarily (see e.g. PR #513). Instead, it would be better to use a `ReaderT` to capture all such parameters.

TerrorJack:
Closed by #525.

PR #525, opened 2020-03-30
Improvements in Asterius.Backends.WasmToolkit

gkaracha:
- Address issue https://github.com/tweag/asterius/issues/522
- Some haddock-compatible comments
- Other minor refactorings


TerrorJack:
There's an OOM on CI when using `wasm-toolkit` with `--no-gc-sections` on the `fib` example. I adjusted the CI script to only test `wasm-toolkit` with ordinary test cases without the `--no-gc-sections` flag for now.

PR #529, opened 2020-03-31
Extend Asterius.CodeGen to marshal atomic operations

gkaracha:
Until now, the `Asterius.CodeGen.marshalCmmPrimCall` would barf (generate a run-time error) if given an atomic `CallishMachOp` [1].

This PR implements code generation for these operations, by translating them non-atomically. Since currently there is no multithreading, this should suffice for now (we might want to revisit this in the future).

[1] That is, one of `AtomicRMW AMO_Add`, `AtomicRMW AMO_Sub`, `AtomicRMW AMO_And`, `AtomicRMW AMO_Nand`, `AtomicRMW AMO_Or`, `AtomicRMW AMO_Xor`, `MO_AtomicRead`, `MO_AtomicWrite`, or `MO_Cmpxchg`.

### DIFF
--- a/asterius/src/Asterius/CodeGen.hs
+++ b/asterius/src/Asterius/CodeGen.hs
@@ -1201,7 +1201,7 @@ marshalCmmPrimCall (GHC.MO_U_QuotRem2 GHC.W64) [q, r] [lhsHi, lhsLo, rhs] = do
         }
   pure [quotout, remout]
 -- Atomic operations
-marshalCmmPrimCall (GHC.MO_AtomicRMW GHC.W64 amop) [dst] [addr,n] =
+marshalCmmPrimCall (GHC.MO_AtomicRMW GHC.W64 amop) [dst] [addr, n] =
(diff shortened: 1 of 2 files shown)


asterius/src/Asterius/CodeGen.hs
```diff
@@ -1139,74 +1139,192 @@ marshalCmmPrimCall (GHC.MO_U_QuotRem2 GHC.W64) [q, r] [lhsHi, lhsLo, rhs] = do
...
+                        mask32 lhsLor 0,
+                        mask32 rhsr 1,
+                        mask32 rhsr 0,
+                        ConstI32 0
+                      ],
+                    callImportReturnTypes = [I32]
+                  }
+          }
+  pure [quotout, remout]
+-- Atomic operations
```
TerrorJack: The diffs before this line seem to be formatting noise. Can you reset the parts above?

gkaracha: Yeah, I overlooked the grouping by `git add --patch`, it should be fixed now.

Issue #515, opened 2020-03-27
Use a Reader monad in Asterius.Backends.Binaryen


gkaracha:
In `Asterius.Backends.Binaryen.hs` we pass several read-only arguments to each marshaling function explicitly (currently the symbol map and a module reference). Thus, extending this list of parameters means that we have to touch a lot of code unnecessarily (see e.g. PR https://github.com/tweag/asterius/pull/513). Instead, it would be better to use a `ReaderT` to capture all such parameters.

gkaracha:
Preliminary attempt to address this shows that it is blocked by Issue https://github.com/tweag/asterius/issues/514, so I will address that one first.

gkaracha:
Closed by https://github.com/tweag/asterius/pull/531.

PR #531, opened 2020-03-31
Minor improvements in Asterius.Backends.Binaryen

gkaracha:
Refactor Asterius.Backends.Binaryen to use a reader monad instead of manually passing all parameters (addresses https://github.com/tweag/asterius/issues/515). The patch does not seem to reduce the size of the file, but hopefully will do so in the future.

### DIFF
--- a/asterius/src/Asterius/Backends/Binaryen.hs
+++ b/asterius/src/Asterius/Backends/Binaryen.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StrictData #-}
 {-# OPTIONS_GHC -Wno-overflowed-literals #-}
 
 -- |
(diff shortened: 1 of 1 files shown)


asterius/src/Asterius/Backends/Binaryen.hs
```diff
@@ -207,164 +217,219 @@ marshalFunctionType FunctionType {..} = flip runContT pure $ do
...
-  BinaryenModuleRef ->
-  Expression ->
-  IO BinaryenExpressionRef
-marshalExpression tail_calls sym_map m e = flip runContT pure $ case e of
+-- TODO: there is a similar definition in Asterius.Backends.WasmToolkit. Maybe
+-- we should just move shared stuff into Asterius.Types (or another shared file).
+type SymbolMap = M.Map AsteriusEntitySymbol Int64
+
+-- | Environment used during marshaling of Asterius' types to Binaryen.
+data MarshalEnv
```
TerrorJack: Since we're adding a new datatype in this module, I suggest adding the `StrictData` extension here, since we don't use laziness in any field; `StrictData` is a good default to get a bit of speed up and plug in some leaks imo.

gkaracha: Good point :+1:

Issue #331, opened 2019-10-20
Make use of precomputed `Int` and `Char` closures in the runtime


TerrorJack:
The ghc rts has precomputed `Int` and `Char` closures for small `Int` and `Char` values. When the garbage collector finds an `Int` or `Char` closure in the dynamic memory which fall into the range, the closure is replaced with a precomputed closure. This trick is based on the assumption that a lot of such closures actually contain small values, and can slightly improve the gc efficiency. We don't implement this trick for now, but it shouldn't be too hard to do so; additionally, `rts_mk*` functions can also return precomputed closures directly.

gkaracha:
Closed by https://github.com/tweag/asterius/pull/534.

TerrorJack:
Let's keep this one open for now. `rts_mkInt`/`rts_mkChar` functions can return precomputed closures now, but the garbage collector doesn't use them yet.

PR #534, opened 2020-04-01
Make use of precomputed `Int` and `Char` closures in the runtime

gkaracha:
This PR addresses https://github.com/tweag/asterius/issues/331.

### DIFF
--- a/asterius/src/Asterius/Builtins.hs
+++ b/asterius/src/Asterius/Builtins.hs
@@ -1162,14 +1162,48 @@ rtsMkCharFunction :: BuiltinsOptions -> AsteriusModule
 rtsMkCharFunction _ = runEDSL "rts_mkChar" $ do
   setReturnTypes [I64]
   [i] <- params [I64]
-  p <- call' "allocate" [mainCapability, constI64 2] I64
-  storeI64 p 0 $ symbol "ghczmprim_GHCziTypes_Czh_con_info"
-  storeI64 p 8 i
-  emit p
+  if'
+    [I64]
+    (i `ltUInt64` constI64 256)
+    -- If the character in question is in the range [0..255] we use the
+    -- trick that GHC uses, and instead of generating a heap-allocated Char
+    -- closure, we simply return the address of the statically allocated
+    -- Char. See stg_CHARLIKE_closure in
+    -- ghc-toolkit/boot-libs/rts/StgMiscClosures.cmm
+    -- Case 1: Use static closure
+    (let offset = i `mulInt64` constI64 16
+     in  emit $ symbol "stg_CHARLIKE_closure" `addInt64` offset
+    )
+    -- Otherwise, we fall back to the more inefficient
+    -- approach and generate a dynamic closure.
+    $ do
+        p <- call' "allocate" [mainCapability, constI64 2] I64
+        storeI64 p 0 $ symbol "ghczmprim_GHCziTypes_Czh_con_info"
+        storeI64 p 8 i
+        emit p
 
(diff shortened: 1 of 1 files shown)


asterius/src/Asterius/Builtins.hs
```diff
@@ -1162,14 +1162,47 @@ rtsMkCharFunction :: BuiltinsOptions -> AsteriusModule
...
 
 rtsMkIntFunction :: BuiltinsOptions -> AsteriusModule
-rtsMkIntFunction opts =
-  rtsMkHelper opts "rts_mkInt" "ghczmprim_GHCziTypes_Izh_con_info"
+rtsMkIntFunction _ = runEDSL "rts_mkInt" $ do
+  setReturnTypes [I64]
+  [i] <- params [I64]
+  if'
+    [I64]
+    ((i `leSInt64` constI64 16) `andInt32` (i `geSInt64` constI64 0xFFFFFFFFFFFFFFF0))
```
TerrorJack, line 1190: IMHO `constI64 (-16)` is better here. In the IR, `ConstI64` is represented as `Int64` which is signed, and it's safe to use a negative literal here for better readability.

asterius/src/Asterius/Builtins.hs
```diff
@@ -1162,14 +1162,47 @@ rtsMkCharFunction :: BuiltinsOptions -> AsteriusModule
...
+    -- If the integer in question is in the range [-16..16] we use the
+    -- trick that GHC uses, and instead of generating a heap-allocated Int
+    -- closure, we simply return the address of the statically allocated
+    -- Int. See stg_INTLIKE_closure in
+    -- ghc-toolkit/boot-libs/rts/StgMiscClosures.cmm
+    ( let offset = (i `addInt64` constI64 16) `mulInt64` constI64 16
```
TerrorJack, line 1196: The correct offset should be `(i+32)*16`.

gkaracha: but `i` is in [-16..16] so `(i+16)` is in [0..32], no?

TerrorJack: Think of this as:

* `stg_INTLIKE_closure`: starting address of the whole segment
* `stg_INTLIKE_closure + 32`: the address of `I# 0#` closure, since there are 16 negative elements in the beginning, each occupying 16 bytes
* `stg_INTLIKE_closure + 32 + (i * 16)`: the address of `I# i` closure

gkaracha: The way think of this is:
`- 16`  in location `stg_INTLIKE_closure + (-16 + 16) * 16` (== `stg_INTLIKE_closure`)
`- 15`  in location `stg_INTLIKE_closure + (-15 + 16) * 16` (== `stg_INTLIKE_closure + 16`)
`- 14`  in location `stg_INTLIKE_closure + (-14 + 16) * 16` (== `stg_INTLIKE_closure + 2*16`)
...
`i` in location `stg_INTLIKE_closure + (i + 16) * 16` 
so basically 16 per piece, indexing starts at zero. I still can't see why this is wrong :/

gkaracha: Ah, I think you made a typo: `I# 0#` is at `stg_INTLIKE_closure + 256` (16 * 16).
With this approach, our numbers end up identical:
`256 + 16*i = 16 * (16 + i)`

TerrorJack: Ah yes, my formula should really be `(i*16)+256`, which is the same as yours :)

Issue #532, opened 2020-04-01
Implement support for `--low-memory-unused` in `binaryen` optimizer


TerrorJack:
The `binaryen` optimizer supports the `--low-memory-unused` argument: given the assumption that the first 1K linear memory is unused, more aggressive optimization can be performed. For improved performance, we should adjust our linker to leave the first 1K memory space to be blank, and switch on this flag.

Relevant places in the source code to look at:

* The `binaryen` C API to get/set the `--low-memory-unused` flag: https://github.com/WebAssembly/binaryen/blob/a32102c7dd7b321330c6cce4d0e3b16e7187a007/src/binaryen-c.h#L1322
* How the Asterius linker sets the (64-bit virtual) starting address of all data segments: https://github.com/tweag/asterius/blob/219e6281207bf882a8c26e46c5ae76ee20b2e419/asterius/src/Asterius/Resolve.hs#L193
* The `binaryen` PR which implemented `--low-memory-unused`: https://github.com/WebAssembly/binaryen/pull/1924

TerrorJack:
Closed by #538.

PR #538, opened 2020-04-02
Unconditionally set --low-memory-unused

gkaracha:
This patch adjusts the linker to leave the first 1KB memory space empty, and enables the `--low-memory-unused` flag (addressing https://github.com/tweag/asterius/issues/532).


asterius/src/Asterius/Main.hs
```diff
@@ -303,6 +303,7 @@ ahcDistMain logger task (final_m, report) = do
       Binaryen.c_BinaryenSetDebugInfo $ if verboseErr task then 1 else 0
       Binaryen.c_BinaryenSetOptimizeLevel $ fromIntegral $ optimizeLevel task
       Binaryen.c_BinaryenSetShrinkLevel $ fromIntegral $ shrinkLevel task
+      Binaryen.c_BinaryenSetLowMemoryUnused $ 1
```
TerrorJack, line 306: If `hlint` is applied here, it would complain for a redundant `$` usage.

gkaracha: Should be ok now

Issue #373, opened 2019-12-05
Missing cbits for `primitive`


TerrorJack:
`primitive` has its own set of functions for `memcpy`, `memmove`, `memcmp` and `memset`. Since it's used by `vector`, maybe this has a high impact on downstream dependencies.

TerrorJack:
Places in the source code to look at:

* The cbits of `primitive` package: https://github.com/haskell/primitive/blob/master/cbits/primitive-memops.h
* How `primitive` imports the cbits: https://github.com/haskell/primitive/blob/master/Data/Primitive/Internal/Operations.hs
* How `primitive` generates the `Prim` instances for various types: https://github.com/haskell/primitive/blob/master/Data/Primitive/Types.hs

gkaracha:
@TerrorJack thanks for the resources. I am still unclear though as to what does this issue require to be done in asterius, can you clarify?

TerrorJack:
Closed by #542.

PR #542, opened 2020-04-04
WIP Address #373

gkaracha:
@TerrorJack this is not ready for reviewing; I'll let you know when it's time. Since addressing https://github.com/tweag/asterius/issues/373 is not as self-contained as the previous ones I have addressed (it needs some more knowledge about asterius' pipeline and runtime), I'll just keep my progress with it visible here for ease.


asterius/src/Asterius/Builtins/Primitive.hs
```diff
@@ -0,0 +1,183 @@
...
+    <> primitiveMemsetFloat
+    <> primitiveMemsetDouble
+    <> primitiveMemsetChar
+
+-- -------------------------------------------------------------------------
+
+-- | @void hsprimitive_memcpy(void *dst, ptrdiff_t doff, void *src, ptrdiff_t soff, size_t len)@
+primitiveMemcpy :: AsteriusModule
+primitiveMemcpy = runEDSL "hsprimitive_memcpy" $ do
+  setReturnTypes []
```
TerrorJack, line 88: `setReturnTypes` only needs to be called when there's actual return value.

asterius/src/Asterius/Builtins/Primitive.hs
```diff
@@ -0,0 +1,183 @@
...
+  [dst, doff, src, soff, len] <- params [I64, I64, I64, I64, I64]
+  let arg1 = dst `addInt64` doff
+      arg2 = src `addInt64` soff
+  callImport "__asterius_Memory_memcpy" $
+    map convertSInt64ToFloat64 [arg1, arg2, len]
```
TerrorJack, line 93: In such cases, we're sure that both the addresses and the length values are non-negative, so it would be better to use `convertUInt64ToFloat64` instead.

gkaracha: Ah right :+1: I changed all calls accordingly (only the value passed to memset now uses `convertSInt64ToFloat64`).

asterius/src/Asterius/Builtins/Primitive.hs
```diff
@@ -0,0 +1,187 @@
...
+
+-- | @void hsprimitive_memset_XXX (XXX *p, ptrdiff_t off, size_t n, XXX x)@
+mkPrimitiveMemsetFloat ::
+  -- | Size (in bytes) of the type
+  Int ->
+  -- | String representation of the type
+  AsteriusEntitySymbol ->
+  AsteriusModule
+mkPrimitiveMemsetFloat size typerep = runEDSL hsname $ do
+  [p, off, n, x] <- params [I64, I64, I64, I64]
```
TerrorJack, line 141: The forth argument type is actually `F32` or `F64`. When compiling a `foreign import ccall`, we treat the unsigned/signed integer types in the argument types/result types all as 64-bit integers, but `Float` indeed maps to `F32` and `Double` indeed maps to `F64`.

asterius/src/Asterius/Builtins/Primitive.hs
```diff
@@ -0,0 +1,187 @@
...
+  -- | Size (in bytes) of the type
+  Int ->
+  -- | String representation of the type
+  AsteriusEntitySymbol ->
+  AsteriusModule
+mkPrimitiveMemsetUInt size typerep = runEDSL hsname $ do
+  [p, off, n, x] <- params [I64, I64, I64, I64]
+  callImport "__asterius_Memory_memset" $
+    [ convertUInt64ToFloat64 $ p `addInt64` off,
+      convertSInt64ToFloat64 x,
```
TerrorJack, line 126: This should be `convertUInt64ToFloat64` as well. Since `x` is the element to be filled in the typed array, and all our typed arrays are created as unsigned versions

gkaracha: I see. Does this apply to the `x` in `mkPrimitiveMemsetFloat` too?

gkaracha: Ah, no it doesn't, I see it now. Ignore my comment

asterius/src/Asterius/Builtins/Primitive.hs
gkaracha: I had a go at it, please have a look (I changed the imports as well to reflect the change).

asterius/rts/rts.memory.mjs
```diff
@@ -299,8 +299,26 @@ export class Memory {
...
+      1 : Uint8Array,
+      2 : Uint16Array,
+      4 : Uint32Array,
+      8 : BigUint64Array
+    };
+    const buf = this.expose(_dst, n, ty[size]);
+    buf.fill(c);
+  }
+
+  memsetFloat32(_dst, c, n) {
```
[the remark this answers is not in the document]
TerrorJack: But then, of course, we would need an extra `size` argument to tell the JS to pick `Float32Array` or `Float64Array`. So this one is a weak suggestion, current form would also work.

asterius/test/primitive/primitive.hs
```diff
@@ -0,0 +1,166 @@
+{-# LANGUAGE BangPatterns #-}
```
TerrorJack, line 1: I assume the unit test is just a WIP? The `primitive` stuff is not called at all and the pasted `fib` logic is not cleaned up.

asterius/rts/rts.memory.mjs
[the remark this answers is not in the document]
gkaracha: Exactly. I thought that it's only two cases so we might as well separate the functions. At least they are small and clear this way. If you agree, I'd resolve this.

asterius/test/primitive/primitive.hs
gkaracha: Yeah, sorry about that, I had already committed locally so it came along :woman_facepalming:

gkaracha:
@TerrorJack: I have been adding some unit tests for the `memset` implementations (see commits 791fae6, c903a30 and 6259119) and I have encountered a problem: running ```stack test asterius:primitive``` shows that

* In `memset(_dst, c, n, size = 1)` `c` is  numeric type, not a `BigInt`; my original implementation gave `TypeError: Cannot convert 0 to a BigInt` when calling memset with `size = 4`.
* Converting `c` explicitly to a `BigInt` does not fix the problem; it makes the type error go away, but we lose bits (e.g. we store `0x00d0cccccccccccc` instead of `0xcccccccccccccccc`).

Have you encountered anything similar before?

PS. `Word8`, `Word16`, `Word32`, `Char`, `Float`, and `Double` all work as expected.

gkaracha:
@TerrorJack I wrapped up `test/primitive` (I even left two tests that print `False`; they shall print `True` when we figure out a way to fix the lossy conversion of `Number` to `BigInt`). This PR should be ready for review now :slightly_smiling_face:

asterius/src/Asterius/Builtins/Primitive.hs
```diff
@@ -0,0 +1,181 @@
...
+      externalModuleName = ext_mod_name,
+      externalBaseName = ext_base_name,
+      functionType = fn_type
+    }
+{-# INLINE mkImport #-}
+
+-- -------------------------------------------------------------------------
+
+-- GEORGE: Note that Asterius uses 64 bits everywhere, but WebAssembly at the
+-- moment does not support I64. Yet, it DOES support F64. Hence, we convert
```
ggreif, line 44: This needs clarification. [WebAssembly *does* support `I64`](https://webassembly.github.io/spec/core/syntax/types.html), and `F64` is subject to floating-point normalisation, so certain values might be problematic.

gkaracha: @ggreif You are right that certain values are problematic because of the representation (see https://github.com/tweag/asterius/blob/df1f5b1ba57619cc14369554eef2a323c0e9cef4/asterius/rts/rts.memory.mjs#L294-L313). We have no clear solution to the problem yet though, that's why I added these comment, lest we forget.

ggreif: I think I get what you mean, I keep forgetting your boundary condition, that you _always_ interface with some JS module. I live in a somewhat different world, where I can basically use all Wasm-permitted types when calling imports, also `I64`.

gkaracha: > I live in a somewhat different world, where I can basically use all Wasm-permitted types when calling imports, also `I64`.

Indeed, I wish I could live in that world too! For now though you are right that this needed more attention. I opened a separate issue documenting this behavior (https://github.com/tweag/asterius/issues/682) instead of keeping this only in comments. Hopefully that clarifies things for others too :slightly_smiling_face:

PR #549, opened 2020-04-08
Use Binary in GHC API for serialization

TerrorJack:
Following #548, we use `Binary` in GHC API for serialization of Asterius IR. Previously we used `binary`, which is easy to implement since `binary` is context-less, but as a consequence, there is no string-consing in the serialized objects. `Binary` supports serializing `FastString`s and `Name`s using symbol tables to dedupe them, so it makes sense to use `Binary` for Asterius IR. This should also result in smaller object files.

To avoid needing to handwrite `Binary` instances for everything, it's possible to write TH or `Generic` logic to generate the logic. The current approach is `Generic`, but the sum type case hasn't been well optimized yet.


TerrorJack:
Closing this one in favor of #553.

PR #550, opened 2020-04-09
Remove obsolete LinkingSection from wasm-toolkit

gkaracha:
@TerrorJack are there any plans to use the other custom section (`RelocationSection`) or shall I remove that one as well?


TerrorJack:
@gkaracha The `RelocationSection` logic shall be removed as well. For now we only need opaque custom sections whose content is a piece of raw bytes without any attached semantics.

gkaracha:
@TerrorJack this PR should be ready for reviewing now.

PR #558, opened 2020-04-12
Increase parallelism in the boot script

TerrorJack:
This PR speeds up the boot process by squashing build commands for multiple boot packages so that we can benefit from increased package-level parallelism. It turns out that `cabal` supports specifying multiple local/hackage targets within one `v1-install` command, so here we go.

`template-haskell` still needs to be built using the `act-as-setup` method. Seems to be a hard limit of `cabal`.


TerrorJack:
Closing this one for now since the TH runner seems to regress.

PR #562, opened 2020-04-13
Use the simplified `binaryen` package

TerrorJack:
This PR uses the simplified `binaryen` package which doesn't bundle the `binaryen` C++ sources and relies on the system to provide the dependency.


TerrorJack:
Closed in favor of #571.

Issue #570, opened 2020-04-14
Use `Bag`s to replace `DList`s


TerrorJack:
**Is your feature request related to a problem? Please describe.**
We rolled a simple `DList` implementation and used it in a few places (e.g. EDSL, wasm-toolkit backend, etc). In fact, this `DList` implementation can be replaced by `Bag` in GHC API; `Bag` doesn't use church encoding to fuse the concat operations, instead it simply represents concat as a new datacon, and this is more efficient in terms of closure representation.

**Describe the solution you'd like**
Remove `DList` logic in the codebase, and use `Bag` in GHC API instead.

PR #572, opened 2020-04-14
[WIP] Address issue #570

gkaracha:
This PR addresses issue https://github.com/tweag/asterius/issues/570, replacing all uses of `DList` with `Bag` to improve performance. @TerrorJack two concerns I have:

1. I don't think that keeping things in order is part of the `Bag` API, which means that it might not have the wanted semantics. This worries me a little; the internal implementation could be changed in the future to reorder things.

2. Surprisingly, `Bag` is not an instance of `Semigroup` so the code can be slightly more verbose in some cases (see for example the changes in `Asterius.Backends.WasmToolkit`).


TerrorJack:
`Bag` is not frequently changing in upstream GHC; even if the ordered sequence semantics is taken away (which is unlikely), we can still roll our own.

mboes:
Have we quantified the performance impact?

TerrorJack:
For `ahc-link` on the `WebOrmolu` example using the `wasm-toolkit` backend, total allocation has dropped from `41,840,454,592 bytes` to `40,443,297,088 bytes`; total time from `46282 ticks` to `42000 ticks`. The number is almost the same for the `binaryen` backend. The performance improvement is significant in `wasm-toolkit` backend since we used `DList` to build instruction sequences there.

To measure the performance impact of a PR, we need to build a profiled version of `asterius`, compile some dependencies (up to `ormolu` in this case), and then link an example with the profiling RTS flags on. This has to be done two times for both the PR and original `master` to diff the profiling reports. This is very time consuming and RAM consuming, so, for now, I manually do this before approving performance PRs.

PR #583, opened 2020-04-15
Use setImmediate.js polyfill instead of setTimeout

pimlu:
Not sure if this is exactly what you want to do about it, but I used this to make my Asterius browser build roughly 2x faster.

This PR is roughly a copy-paste of [setImmediate.js](https://github.com/YuzuJS/setImmediate).

However there are some minor modifications to the polyfill's `global`/`attachTo` logic to avoid writing the polyfill to the global scope.


TerrorJack:
The nodejs platform seems to be broken by this. But anyway, thanks for the input that `setImmediate.js` makes things faster by 2x in the browser build!

PR #585, opened 2020-04-16
[WIP] Address issue 569

gkaracha:


gkaracha:
@TerrorJack, this should be ready for review now. As per your feedback, I used `IntMap` directly (instead of `UniqFM`), and kept the `containers` naming convention (instead of the GHC-internals one).

asterius/src/Asterius/Types/SymbolMap.hs
```diff
@@ -0,0 +1,239 @@
...
+newtype SymbolMap a = SymbolMap (IM.IntMap (EntitySymbol, a))
+  deriving newtype (Eq)
+  deriving stock (Data)
+
+instance Show a => Show (SymbolMap a) where
+  showsPrec d m =
+    showParen (d > 10) $
+      showString "fromList " . shows (toListSM m)
+
+instance Semigroup (SymbolMap a) where
```
TerrorJack, line 86: The `Semigroup`/`Monoid` instances also belong to `deriving newtype`.

asterius/src/Asterius/Types/SymbolMap.hs
```diff
@@ -0,0 +1,239 @@
...
+    fromMap,
+  )
+where
+
+import Asterius.Binary.Orphans ()
+import Asterius.Types.EntitySymbol
+import Binary
+import Control.Monad
+import Data.Data
+import qualified Data.IntMap as IM
```
TerrorJack, line 67: Better be explicitly importing `Data.IntMap.Lazy` here. The `Lazy` and `Strict` versions are the same type, and `Data.IntMap` currently point to the lazy version, but still better explicit than implicit.

TerrorJack: The same applies to `Data.Map`.

asterius/src/Asterius/Types/SymbolMap.hs
```diff
@@ -0,0 +1,239 @@
...
+
+instance Semigroup (SymbolMap a) where
+  SymbolMap m1 <> SymbolMap m2 = SymbolMap $ m1 <> m2
+
+instance Monoid (SymbolMap a) where
+  mempty = empty
+  mappend = (<>)
+
+instance Binary a => Binary (SymbolMap a) where
+  -- We are lazy on both at the moment, hence
```
TerrorJack, line 94: Why are we lazy on both?

asterius/src/Asterius/Types/SymbolMap.hs
```diff
@@ -0,0 +1,239 @@
...
+
+-- | /O(1)/. The empty map.
+{-# INLINE empty #-}
+empty :: SymbolMap a
+empty = SymbolMap IM.empty
+
+-- | /O(1)/. A map containing a single association.
+{-# INLINE singleton #-}
+singleton :: EntitySymbol -> a -> SymbolMap a
+singleton k e = SymbolMap $ IM.singleton (getKey $ getUnique k) (k, e)
```
TerrorJack, line 123: There're multiple `getKey $ getUnique k` usages in this module, better factor it out to an inline function.

asterius/src/Asterius/Types/SymbolMap.hs
```diff
@@ -0,0 +1,239 @@
...
+-- | /O(min(n,W))/. Lookup the value at an 'EntitySymbol' in the map.
+lookup :: EntitySymbol -> SymbolMap a -> Maybe a
+lookup k (SymbolMap m) = snd <$> IM.lookup (getKey $ getUnique k) m
+
+-- | /O(min(n,W))/. Find the value at an 'EntitySymbol'. Calls 'error' when the
+-- element can not be found.
+(!) :: HasCallStack => SymbolMap a -> EntitySymbol -> a
+(!) m k = case lookup k m of
+  Just e -> e
+  Nothing -> error "SymbolMap.!: given key is not an element in the map"
```
TerrorJack, line 149: Better report the missing key here.

asterius/src/Asterius/Types/SymbolMap.hs
```diff
@@ -0,0 +1,239 @@
...
+mapAccum :: (a -> b -> (a, c)) -> a -> SymbolMap b -> (a, SymbolMap c)
+mapAccum f z (SymbolMap m) =
+  SymbolMap <$> IM.mapAccum (\a (k, e) -> fmap (k,) $ f a e) z m
+
+-- | /O(n)/. Fold the keys and values in the map using the given
+-- right-associative binary operator. This is a strict variant: each
+-- application of the operator is evaluated before using the result in the next
+-- application. This function is strict in the starting value.
+foldrWithKey' :: (EntitySymbol -> a -> b -> b) -> b -> SymbolMap a -> b
+foldrWithKey' f z (SymbolMap m) = IM.foldrWithKey (\_ (k, e) b -> f k e b) z m
```
TerrorJack, line 198: We should be using `IM.foldrWithKey'` here.

asterius/src/Asterius/Types/SymbolMap.hs
gkaracha: Ah, about that. The `Key` we are strict on, but the contents are `(EntitySymbol, a)`. But we serialize as if it were a `Map EntitySymbol a`, so only the (lazy) contents are used, the keys are ignored completely.

asterius/src/Asterius/Types/SymbolMap.hs
gkaracha: Done (I created `getKeyES`).

asterius/src/Asterius/Types/SymbolMap.hs
gkaracha: Done.

asterius/src/Asterius/Types/SymbolMap.hs
gkaracha: Ah, good catch! Fixed :slightly_smiling_face:

asterius/src/Asterius/Types/SymbolMap.hs
gkaracha: > better explicit than implicit.

I definitely agree. Should be OK now.

asterius/src/Asterius/Types/SymbolMap.hs
gkaracha: Done.

asterius/src/Asterius/Types/SymbolMap.hs
TerrorJack: The `EntitySymbol` is guaranteed to be already evaluated. In the tuple, only the `a` field is lazy. So we should use strict `put_` and `get` for `EntitySymbol`.

gkaracha: I see. Should be OK now.

TerrorJack:
Confirmed performance improvement for `ahc-link`: `73074 ticks` to `67507 ticks`. Less obvious for `ahc-ld`: `65004 ticks` to `64993 ticks`.

Issue #565, opened 2020-04-14
Get rid of the `AsteriusModuleSymbol` type


TerrorJack:
**Is your feature request related to a problem? Please describe.**
`AsteriusModuleSymbol` was the clone of `Module` in GHC API. We used to include it in the parts of serialized Asterius IR, but now it's only used as a compile-time type for indexing in certain places. We can get rid of this type and use vanilla `Module` to reduce code and avoid some marshaling overhead.

**Describe the solution you'd like**
Remove `AsteriusModuleSymbol` definition and usages everywhere, use `Module` in GHC API instead.

PR #589, opened 2020-04-17
WIP Fix Issue #565

gkaracha:
Closes https://github.com/tweag/asterius/issues/565.

### DIFF
--- a/asterius/src/Asterius/TypesConv.hs
+++ b/asterius/src/Asterius/TypesConv.hs
@@ -3,8 +3,7 @@
 {-# LANGUAGE RecordWildCards #-}
 
 module Asterius.TypesConv
-  ( marshalToModuleSymbol,
-    zEncodeModule,
+  ( zEncodeModule,
     generateWasmFunctionTypeSet,
     asmPpr,
     asmPrint,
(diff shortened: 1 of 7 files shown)


asterius/src/Asterius/TypesConv.hs
```diff
@@ -22,23 +21,19 @@ import System.IO
...
-  { unitId = GHC.fs_bs $ GHC.unitIdFS u,
-    moduleName =
-      CBS.splitWith (== '.') $ GHC.fs_bs $ GHC.moduleNameFS m
-  }
-
-{-# INLINE zEncodeModuleSymbol #-}
-zEncodeModuleSymbol :: AsteriusModuleSymbol -> String
-zEncodeModuleSymbol AsteriusModuleSymbol {..} =
+{-# INLINE zEncodeModule #-}
+zEncodeModule :: GHC.Module -> String
```
TerrorJack, line 25: We can obtain the unit id part as `unitId = GHC.unitIdFS m`, module name part as `moduleName = GHC.moduleNameFS m`, so we can simply concat these two as `FastString`: `unitId <> "_" <> moduleName`, avoiding redundant computations.

gkaracha: Ah, right. I was too hasty with this but it should be fixed now.

Issue #586, opened 2020-04-17
Use hashsets to replace `Set`s


TerrorJack:
**Is your feature request related to a problem? Please describe.**
Similar to #569, we used to use `Set EntitySymbol` in various places (particularly the gc-sections pass in `Asterius.Passes.GCSections`, where there are a lot of calculations about symbol sets). It's possible to use hashsets to replace `Set`s for better performance.

**Describe the solution you'd like**
Implement the `SymbolSet` type, using the `Uniquable` instance of `EntitySymbol` to calculate the hash value, and the `IntSet` type under the hood. (iirc we don't need to recover the actual `EntitySymbol`s, since there's no need for `toList`).

PR #590, opened 2020-04-17
WIP Fix Issue #586

gkaracha:
Closes https://github.com/tweag/asterius/issues/586. Still WIP.


gkaracha:
@TerrorJack, I wonder whether there is a neat solution to https://github.com/tweag/asterius/issues/586 or not. I want to give `foldr'` the expected type signature
```haskell
foldr' :: (EntitySymbol -> b -> b) -> b -> SymbolSet -> b
```
but that is literally impossible because `EntitySymbol` appears in a negative position and the `SymbolSet` does not really contain `EntitySymbol`s but their keys (and `getKeyES` is a one-way function). I was tempted to use the other option:
```haskell
foldr' :: (Key -> b -> b) -> b -> SymbolSet -> b
```
but this means that we need to change the call sites and the implementation starts leaking. :confused: Even worse, the call site (`Asterius.Passes.GCSections.gcSections`) seems to really need to use the complete `EntitySymbol`, so having it around seems unavoidable.

I am afraid that we are better where we are already. What do you think?

TerrorJack:
So we still need to remember `EntitySymbol` in the `SymbolSet`. We can do `IntMap EntitySymbol`; it'll still be faster than `Set EntitySymbol`.

gkaracha:
> it'll still be faster than `Set EntitySymbol`.

Ha, you think so? I gave it a shot, let me know how it goes! :slightly_smiling_face:

asterius/src/Asterius/Types/SymbolSet.hs
```diff
@@ -0,0 +1,123 @@
...
+    -- ** Lists
+    fromList,
+    toList,
+  )
+where
+
+import Asterius.Types.EntitySymbol
+import Data.Data
+import qualified Data.Foldable as Foldable
+import qualified Data.IntMap.Lazy as IM
```
TerrorJack, line 46: We should be using `Data.IntMap.Strict` for `SymbolSet`, since the symbols are never lazy.

asterius/src/Asterius/Types/SymbolSet.hs
```diff
@@ -0,0 +1,123 @@
...
+empty = SymbolSet IM.empty
+
+-- | /O(1)/. A set of one element.
+{-# INLINE singleton #-}
+singleton :: EntitySymbol -> SymbolSet
+singleton e = SymbolSet $ IM.singleton (getKeyES e) e
+
+-- | /O(n+m)/. The union of two sets.
+{-# INLINE union #-}
+union :: SymbolSet -> SymbolSet -> SymbolSet
```
TerrorJack, line 80: I think we can directly `coerce IM.union` here.

TerrorJack: The same may apply to all other functions which simply wraps/unwraps the `newtype` around the `IntMap` and reuses an `IntMap` interface. Always prefer `coerce` if possible.

asterius/src/Asterius/Types/SymbolSet.hs
```diff
@@ -0,0 +1,123 @@
...
+-- is strict in the starting value.
+{-# INLINE foldr' #-}
+foldr' :: (EntitySymbol -> b -> b) -> b -> SymbolSet -> b
+foldr' fn z = IM.foldr' fn z . fromSymbolSet
+
+-- ----------------------------------------------------------------------------
+
+-- | /O(n*min(n,W))/. Convert a 'SymbolSet' to an 'IS.IntSet' (internal use).
+{-# INLINE toIntSet #-}
+toIntSet :: SymbolSet -> IS.IntSet
```
TerrorJack, line 109: Use `keysSet` instead.

gkaracha: :+1:

asterius/src/Asterius/Types/SymbolSet.hs
gkaracha: I did apply your advice to all functions that I could in this file (though I needed to add some explicit type applications to bypass GHC's failing type inference). 

> Always prefer `coerce` if possible.

I see that with `coerce` the inliner gets to inline the functions more eagerly so that's nice. Is there any other benefit in choosing `coerce` over newtype wrapping/unwrapping? I have the impression that it's all the same to GHC; such no-ops are eventually optimized anyway.

asterius/src/Asterius/Types/SymbolSet.hs
gkaracha: Ah, right. Fixed.

asterius/src/Asterius/Types/SymbolSet.hs
TerrorJack: IMO it's mainly about code brevity.

gkaracha:
@TerrorJack, did this PR lead to an improvement after all?

TerrorJack:
@gkaracha 
Before: https://buildkite.com/tweag-1/asterius/builds/360#6547fe22-bb2e-43fb-a034-aa9e28227804
After: https://buildkite.com/tweag-1/asterius/builds/362#c21a5b96-b4de-4c80-9be2-78f032e29155
You can check the `ahc-ld.prof` file in the artifacts to check the difference.

gkaracha:
Thanks! :slightly_smiling_face: Pasting here the difference, for future reference:

```diff
<       total time  =     1104.97 secs   (1104974 ticks @ 1000 us, 1 processor)
<       total alloc = 145,237,340,288 bytes  (excludes profiling overheads)
---
>       total time  =     1010.16 secs   (1010158 ticks @ 1000 us, 1 processor)
>       total alloc = 143,418,583,240 bytes  (excludes profiling overheads)
```

Issue #592, opened 2020-04-18
Emit smaller `__asterius_unreachable` basic blocks


TerrorJack:
**Is your feature request related to a problem? Please describe.**
When marshaling the CFG in a Cmm function, we need a special `__asterius_unreachable` basic block as the jump target in case a `CmmSwitch` lacks a default target. `__asterius_unreachable` would report a runtime error using `barf` before crashing.

When GHC emits `CmmSwitch` without a default target, it knows the full list of possible values to match (e.g. for a `case` expression), and it's safe to assume `__asterius_unreachable` is never actually reached. And in practice, we haven't seen runtime errors related to `__asterius_unreachable` in the wild.

We should emit `__asterius_unreachable` basic blocks only when necessary, and make them small. This should result in a small binary size improvement in each Cmm function.

**Describe the solution you'd like**
* When building the CFG for a Cmm function, only emit an `__asterius_unreachable` basic block when there is at least one `CmmSwitch` without default target in that function
* Our own relooper for the `wasm-toolkit` backend assumes `__asterius_unreachable` exists, so re-insert it if it's absent from the input CFG
* Each `__asterius_unreachable` only contains one `unreachable` instruction.

PR #602, opened 2020-04-21
Improve Asterius.CodeGen

gkaracha:
This PR closes https://github.com/tweag/asterius/issues/592.

### DIFF
--- a/asterius/src/Asterius/CodeGen.hs
+++ b/asterius/src/Asterius/CodeGen.hs
@@ -1593,7 +1593,7 @@ unreachableRelooperBlock = RelooperBlock
 marshalCmmProc :: GHC.CmmGraph -> CodeGen Function
 marshalCmmProc GHC.CmmGraph {g_graph = GHC.GMany _ body _, ..} = do
   entry_k <- marshalLabel g_entry
-  (_with_def, rbs) <- do
+  (with_def, rbs) <- do
     let fn ::
           ContainsDefault ->
           (GHC.Label, GHC.Block GHC.CmmNode GHC.C GHC.C) ->
@@ -1603,8 +1603,9 @@ marshalCmmProc GHC.CmmGraph {g_graph = GHC.GMany _ body _, ..} = do
           (b, with_def) <- marshalCmmBlock (GHC.blockToList inner_nodes) exit_node
           pure (with_def_acc <> with_def, (k, b))
     mapAccumLM fn mempty (GHC.bodyList body)
-  let blocks_unresolved =
-        ("__asterius_unreachable", unreachableRelooperBlock) : rbs
+  let blocks_unresolved = case with_def of
+        WithDefault -> rbs
+        WithoutDefault -> ("__asterius_unreachable", unreachableRelooperBlock) : rbs
   pure $ adjustLocalRegs Function
     { functionType = FunctionType {paramTypes = [], returnTypes = []},
       varTypes = [],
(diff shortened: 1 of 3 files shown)


asterius/src/Asterius/CodeGen.hs
```diff
@@ -1465,9 +1466,29 @@ marshalCmmInstr instr = case instr of
 marshalCmmBlockBody :: [GHC.CmmNode GHC.O GHC.O] -> CodeGen [Expression]
 marshalCmmBlockBody instrs = concat <$> for instrs marshalCmmInstr
 
+-- ----------------------------------------------------------------------------
+
+-- | Datatype to capture whether there is at least on CmmSwitch without a
+-- DEFAULT target within a Cmm function. TODO: Place appropriately.
+data ContainsDefault = WithDefault | WithoutDefault
```
TerrorJack, line 1473: My brain needs an extra layer of indirection when processing this thing: `WithoutDefault` -> need to insert `__asterius_unreachable`, `WithDefault` -> do not insert `__asterius_unreachable`.

My suggestion is:

* Rename the datatype from `ContainsDefault` to `NeedsUnreachableBlock`.
* Don't roll the `Semigroup`/`Monoid` instances by hand. Make `NeedsUnreachableBlock` a `newtype` of `Bool`, and use `deriving via` to get the instances of `Data.Monoid.Any`. Things are more obvious this way; when combining a bunch of `NeedsUnreachableBlock`s, if any one of them is true, then the result is true.

asterius/src/Asterius/CodeGen.hs
```diff
@@ -1465,9 +1466,29 @@ marshalCmmInstr instr = case instr of
...
+  -- is exhaustive (innocent until proven guilty).
+  WithoutDefault <> _ = WithoutDefault
+  WithDefault <> y = y
+
+-- ----------------------------------------------------------------------------
+
 marshalCmmBlockBranch ::
   GHC.CmmNode GHC.O GHC.C ->
-  CodeGen ([Expression], Maybe Expression, [RelooperAddBranch])
+  CodeGen ([Expression], Maybe (Expression, ContainsDefault), [RelooperAddBranch])
```
TerrorJack, line 1491: `NeedsUnreachableBlock` should be a standalone field in the outer tuple type. We may produce no `Expression` for a Cmm exit node, but there will still be a `NeedsUnreachableBlock` value (which is false)

asterius/src/Asterius/CodeGen.hs
```diff
@@ -1516,59 +1538,74 @@ marshalCmmBlockBranch instr = case instr of
...
+          NeedsUnreachableBlock ->
+          (GHC.Label, GHC.Block GHC.CmmNode GHC.C GHC.C) ->
+          CodeGen (NeedsUnreachableBlock, (BS.ByteString, RelooperBlock))
+        fn needs_unreachable_acc (lbl, GHC.BlockCC _ inner_nodes exit_node) = do
+          k <- marshalLabel lbl
+          (b, needs_unreachable) <- marshalCmmBlock (GHC.blockToList inner_nodes) exit_node
+          pure (needs_unreachable_acc <> needs_unreachable, (k, b))
+    mapAccumLM fn mempty (GHC.bodyList body)
+  let blocks_unresolved
+        | needs_unreachable == coerce True =
```
TerrorJack, line 1605: We can do `| coerce needs_unreachable = ...` here. And thus I think there's no need for `Eq` and `Show` instances for `NeedsUnreachableBlock`

PR #605, opened 2020-04-22
[WIP] Optimize the gcSections pass

gkaracha:


asterius/src/Asterius/Types.hs
```diff
@@ -94,6 +95,32 @@ data AsteriusStatics
       }
   deriving (Show, Data)
 
+-- ----------------------------------------------------------------------------
+
+data AsteriusCachedModule
```
TerrorJack, line 100: I suggest adding `AsteriusModule` as a field of `AsteriusCachedModule`, instead of just duplicating all the fields.

gkaracha: Since `gcSections` calls `collectEntitySymbols` on each `AsteriusStatics` and `Function` within a module, I basically paired each `SymbolSet` with the entity it comes from:
```diff
- staticsMap :: SymbolMap AsteriusStatics,
+ staticsMap :: SymbolMap (AsteriusStatics, SymbolSet),
---
- functionMap :: SymbolMap Function,
+ functionMap :: SymbolMap (Function, SymbolSet),
```
If I understand correctly `gcSections`, the `SymbolSet` for a complete module depends on the `export_funcs`, which are passed dynamically. This means that we cannot precompute `SymbolSet` at the module-level statically. Hence the representation I use does not reuse `AsteriusModule`. Did I miss something? Do you think we can do better than this?

TerrorJack: My suggestion is having an extra `SymbolMap SymbolSet` field in addition to the `AsteriusModule` field. This represents the mapping from a single symbol to a set of symbols, which is the "adjacency list" representation of a directed graph. Thus we can achieve minimal intrusion to the original logic.

gkaracha: Hmmmm, I see what you mean, I'll introduce an `dependencyMap :: SymbolMap SymbolSet` then. FWIW, I dislike it a little that this introduces an implicit invariant that
```haskell
keys staticsMap `union` keys functionMap == keys dependencyMap`
```
but you are right that it is the least intrusive.

gkaracha:
**Preliminary results**

@TerrorJack, preliminary comparison between [HEAD](https://buildkite.com/tweag-1/asterius/builds/427#dcab9dce-6a4f-472e-9688-34a3949b1348) and [master](https://buildkite.com/tweag-1/asterius/builds/410#0cfda00d-a939-4420-85df-880cc3c9b558):

gkaracha:
**Preliminary results**

@TerrorJack, preliminary comparison between [HEAD](https://buildkite.com/tweag-1/asterius/builds/427#dcab9dce-6a4f-472e-9688-34a3949b1348) and [master](https://buildkite.com/tweag-1/asterius/builds/410#0cfda00d-a939-4420-85df-880cc3c9b558):

`diff master/ahc.prof HEAD/ahc.prof` shows:
```diff
<       total time  =        0.24 secs   (245 ticks @ 1000 us, 1 processor)
<       total alloc = 130,608,824 bytes  (excludes profiling overheads)
---
>       total time  =        0.24 secs   (244 ticks @ 1000 us, 1 processor)
>       total alloc = 132,291,880 bytes  (excludes profiling overheads)
```
`diff master/ahc-link.prof HEAD/ahc-link.prof` shows
```diff
<       total time  =       48.00 secs   (47998 ticks @ 1000 us, 1 processor)
<       total alloc = 3,320,867,416 bytes  (excludes profiling overheads)
---
>       total time  =       45.61 secs   (45606 ticks @ 1000 us, 1 processor)
>       total alloc = 3,320,989,152 bytes  (excludes profiling overheads)
```
`diff master/ahc-ld.prof HEAD/ahc-ld.prof` shows:
```diff
<       total time  =       32.17 secs   (32166 ticks @ 1000 us, 1 processor)
<       total alloc = 125,277,473,792 bytes  (excludes profiling overheads)
---
>       total time  =        9.68 secs   (9675 ticks @ 1000 us, 1 processor)
>       total alloc = 111,164,588,648 bytes  (excludes profiling overheads)
```

gkaracha:
@TerrorJack I think it's ready for review now :slightly_smiling_face:

asterius/src/Asterius/GHCi/Internals.hs
```diff
@@ -287,7 +287,9 @@ asteriusWriteIServ hsc_env i a
                 linkOutput = "",
                 linkObjs = [],
                 linkLibs = [],
-                linkModule = m <> M.foldr' (<>) (ghciLibs s) (ghciObjs s),
+                linkModule =
```
TerrorJack, line 290: IMO this will impact TH performance a lot; the `linkModule` field we pass to the linker logic contains all the bundled archives/objects without the symbol graph info. And the linker will reconstruct the symbol graph on the fly, which is wasteful. I can see two different fixes:

* Make `linkModule` an `AsteriusCachedModule` too, and don't throw away the loaded symbol graph by calling `fromCachedModule` here
* Change the `ghciLibs` and `ghciObjs` representation; we no longer store the decoded modules in-memory, instead we'll simply store their file paths, and put the file paths in the `linkObjs`/`linkLibs` field here when calling `linkExeInMemory`.

gkaracha: I see. I'll have a go at the first approach and see how it goes.

gkaracha:
[master](https://buildkite.com/tweag-1/asterius/builds/417#82c38d50-3049-49e5-b337-e60554b5be45) vs [HEAD](https://buildkite.com/tweag-1/asterius/builds/430#75544f90-4075-49af-8255-49b26b56d4c7) gives:
`diff master/ahc.prof HEAD/ahc.prof`:
```diff
<       total time  =        0.25 secs   (246 ticks @ 1000 us, 1 processor)
<       total alloc = 130,610,000 bytes  (excludes profiling overheads)
---
>       total time  =        0.28 secs   (283 ticks @ 1000 us, 1 processor)
>       total alloc = 132,159,648 bytes  (excludes profiling overheads)
```

`diff master/ahc-link.prof HEAD/ahc-link.prof`:
```diff
<       total time  =       45.99 secs   (45991 ticks @ 1000 us, 1 processor)
<       total alloc = 3,320,876,440 bytes  (excludes profiling overheads)
---
>       total time  =       47.77 secs   (47772 ticks @ 1000 us, 1 processor)
>       total alloc = 3,321,019,664 bytes  (excludes profiling overheads)
```

`diff master/ahc-ld.prof HEAD/ahc-ld.prof `:
```diff
<       total time  =       32.59 secs   (32585 ticks @ 1000 us, 1 processor)
<       total alloc = 125,276,180,272 bytes  (excludes profiling overheads)
---
>       total time  =       11.17 secs   (11166 ticks @ 1000 us, 1 processor)
>       total alloc = 111,162,737,872 bytes  (excludes profiling overheads)
```
I think this is good to go @TerrorJack! :slightly_smiling_face:

PR #612, opened 2020-04-24
Fix some typos in docs

gkaracha:


TerrorJack:
What spell checker did you use? Maybe we can add that to the doc building job on CircleCI to warn about typos in the future.

gkaracha:
> What spell checker did you use? Maybe we can add that to the doc building job on CircleCI to warn about typos in the future.

Nah, it was manually done: while reading them I just enabled my vim spellchecker `:set spell spelllang=en_uk`. Would be nice to have some automation indeed, but I wish I knew of any tool to suggest.

PR #622, opened 2020-04-27
Add some parallelization to ahc-ld and ahc-link

gkaracha:
this is WIP


asterius/src/Asterius/Ld.hs
```diff
@@ -45,9 +46,13 @@ data LinkTask
 loadTheWorld :: LinkTask -> IO AsteriusCachedModule
 loadTheWorld LinkTask {..} = do
   ncu <- newNameCacheUpdater
-  lib <- mconcat <$> for linkLibs (loadAr ncu)
-  objs <- rights <$> for linkObjs (tryGetFile ncu)
+  lib <- parallelFor 1 linkLibs (loadAr ncu) -- TODO: Parameterize
```
TerrorJack, line 49: `loadAr` is already parallelized, and here we use `parallelFor` again, which creates nested parallelism. Not to say this won't run, but the overall efficiency will be reduced.

We need to either:

* Refactor `loadAr`, so that it only returns the `ArchiveEntry`s, and after we collect all the `ArchiveEntry`s, parallelize their deserialization (along with `objs`)
* Improve our parallelization framework so that it supports nested parallelism; each job can spawn new jobs to be evaluated somewhere else.

asterius/app/ahc-ld.hs
```diff
@@ -20,6 +20,10 @@ parseLinkTask args = do
         linkObjs = link_objs,
         linkLibs = link_libs,
         linkModule = mempty,
+        poolSize =
```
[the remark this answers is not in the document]
gkaracha: Ah, indeed, that's better :+1: Will do.

asterius/src/Asterius/Internals/Parallel.hs
```diff
@@ -0,0 +1,47 @@
...
+  | n >= 2 = error "TODO"
+  -- If there are not enough resources, fall back to the sequential version.
+  | otherwise = mapM_ fn xs
+
+-- | Given the worker thread pool capacity @c@, @parallelFor c xs f@ maps @f@
+-- on @xs@ in parallel on the global thread pool, and concatenates the results.
+-- If @c = 1@ then it is equivalent to @mconcat <$> mapM fn xs@, thus avoiding
+-- threading overhead.
+{-# INLINE parallelFor #-}
+parallelFor :: Monoid r => Int -> [a] -> (a -> IO r) -> IO r
```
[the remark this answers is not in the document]
gkaracha: > This splits the input to different chunks and distribute them to the worker threads. IMO this has higher risk of unevenly distribute work.

That's true, with this approach we get idle threads sometimes.

> A better approach is using an `IORef [a]` to coordinate between the worker threads, each thread using `atomicModifyIORef'` to dynamically fetch an element and do the computation (and exit once there's no more elements).

Indeed, this wastes no resources, but it also is really concurrent, possibly changing the order of things. `Monoid r` gives us associativity but what you're suggesting also needs symmetry to be correct. Is it the case that `m1 <> m2 == m2 <> m1` for `AsteriusModule`s?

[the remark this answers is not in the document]
TerrorJack: We don't need to worry about ordering of `<>`; we never concat two `AsteriusModule`s which contain the same symbols.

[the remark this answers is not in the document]
gkaracha: Right. I point out the symmetry requirement in the function description anyway (since the function is parametric over `r`), to be fair. So, does bd7bfa6b15c1dc70b764856e8638a788f2062149 seem to implement what you have in mind?

asterius/src/Asterius/Internals/Parallel.hs
```diff
@@ -0,0 +1,35 @@
...
+-- on @xs@ in parallel on the global thread pool, and concatenates the results.
+-- If @c = 1@ then it is equivalent to @mconcat <$> mapM fn xs@, thus avoiding
+-- threading overhead. NOTE: this function does not only assume associativity
+-- for @<>@ (as promised by the @Monoid@ instance), but also symmetry.
+parallelFor :: Monoid r => Int -> [a] -> (a -> IO r) -> IO r
+parallelFor n xs fn
+  | n >= 2 = do
+    input <- newIORef xs
+    mvars <- replicateM n newEmptyMVar
+    let getNextElem = atomicModifyIORef input $ \case
```
TerrorJack, line 23: We should prefer strict `atomicModifyIORef'` here.

asterius/src/Asterius/Internals/Parallel.hs
```diff
@@ -0,0 +1,35 @@
...
+          [] -> ([], Nothing)
+          (y:ys) -> (ys, Just y)
+        loop mvar acc = getNextElem >>= \case
```
TerrorJack, line 26: Don't forget to use `!acc`. We want the accumulation to happen in the worker thread, instead of passing a thunk consisting a long chain of `<>` to the parent thread.

asterius/src/Asterius/Internals/Parallel.hs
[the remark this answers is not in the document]
TerrorJack: See latest comments.

asterius/src/Asterius/Ld.hs
gkaracha: I went with the first approach:

> Refactor `loadAr`, so that it only returns the `ArchiveEntry`s, and after we collect all the `ArchiveEntry`s, parallelize their deserialization (along with `objs`)

but I don't completely understand what you mean with the second bullet:

> Improve our parallelization framework so that it supports nested parallelism; each job can spawn new jobs to be evaluated somewhere else.

Nested parallelism is already allowed, right? I understand that nested calls would pin threads to the same capabilities (hence slower), but it'd still work. What is the alternative solution you mean here?

TerrorJack: Ah, the second approach elaborated is:

* We have a `P` monad which you can lift `IO` actions into. It carries the context (thread pool, task queue) needed for all parallelism
* There's a `fork :: P a -> P (Future a)` method which spawns a `P a` action for parallel execution. The `P a` action itself can `fork` new jobs, thus nested parallelism here.
* There's a `join :: Future a -> P a` method which blocks until the result is available.

Going through all the trouble of creating the `P` monad abstraction has an advantage: we can enforce the number of worker threads to match CPU core count all the time. Directly calling `parallelFor` inside `parallelFor` would be much less efficient.

gkaracha:
@TerrorJack, I think it is ready for reviewing now. I skipped parallelization of data segments in the binaryen backend as we discussed so that it doesn't backfire on us. I have also left a couple of `TODO`s in the comments I'd like you to have a look at if you can.

asterius/src/Asterius/Types.hs
```diff
@@ -126,6 +128,18 @@ instance Semigroup AsteriusModule where
...
+-- | Given the worker thread pool capacity @c@, @parForceAsteriusModule c m@
+-- deeply evaluates an 'AsteriusModule' m in parallel on the global thread
+-- pool.
+parForceAsteriusModule :: Int -> AsteriusModule -> AsteriusModule
+parForceAsteriusModule n m@(AsteriusModule sm se fm spt mod_ffi_state) =
+  parallelRnf n (SM.toList sm)
+    `seq` parallelRnf n (SM.toList se)
+    `seq` parallelRnf n (SM.toList fm)
+    `seq` parallelRnf n (SM.toList spt)
+    `seq` rnf mod_ffi_state -- TODO: is it worth it parallelizing deeper?
```
TerrorJack, line 140: I think not; `mod_ffi_state` should only count for a small part of the whole `AsteriusModule`.

asterius/src/Asterius/Main.hs
```diff
@@ -103,6 +103,11 @@ parseTask args = case err_msgs of
              in if i >= 0 && i <= 2
                   then t {shrinkLevel = i}
                   else error "Shrink level must be [0..2]",
+          str_opt "thread-pool-size" $ \s t ->
```
TerrorJack, line 106: Don't forget to pass the `--thread-pool-size` option to `ahc-ld` as well. `ahc-ld` is called as the linker executable by `ahc`, so search for `--optl` in this module and see how other `ahc-ld` options are being passed.

gkaracha: Good catch! Should be ok now.

TerrorJack:
We have to hold this one for a bit longer; I gave this a try on my build server, and the overall clock time of linking `pandoc` has actually increased with `--thread-pool-size=8` compared to `1`. So we need to investigate further:

* Are we picking the right places for parallelization
* Are we implementing `parallelFor` correctly

Merely reading the flamegraph doesn't seem to help too much either. I think we also need to investigate the eventlogs with tools like `threadscope` and such.

gkaracha:
@TerrorJack I took your advice and tried out `threadscope`:

> * Are we implementing `parallelFor` correctly

The implementation of `parallelFor` was actually not OK; laziness got in the way. Hopefully this is now fixed (see late commit https://github.com/tweag/asterius/pull/622/commits/06efdb99208f938502a8be6689e4b50c39994194).

> * Are we picking the right places for parallelization

Not sure about that, but if you could have another go at linking `pandoc` we might find out more about this.

TerrorJack:
Nice finding about the missed laziness in `parallelFor`. Do you have updated numbers for the `Cabal` test case locally, for thread pool sizes `1`, `2`, `4`, `8`?

gkaracha:
@TerrorJack some updates on this one:

Inspired by your fix for https://github.com/tweag/asterius/pull/663 (https://github.com/tweag/asterius/commit/0085dcf4c7b3200362092efbf8c7cdc751b5d3db), I tried giving up on parallelization in `loadTheWorld` (https://github.com/tweag/asterius/commit/56a5aeb97b771ebea3cad6884ebf59253c176155), which basically combined all the object files into one eagerly. Indeed, this took the OOM errors away, but it also showed no benefit whatsoever in the parallelization of the rest. For
```haskell
import Distribution.Simple
main = defaultMain
```
wall clock time seems to only be getting worse:
```
$ time stack build --exec "ahc-link --thread-pool-size=1 --input-hs perf-test/Setup.hs"
real    0m46,915s
user    3m56,870s
sys     0m2,098s

$ time stack build --exec "ahc-link --thread-pool-size=2 --input-hs perf-test/Setup.hs"
real    0m53,966s
user    4m2,502s
sys     0m2,822s

$ time stack build --exec "ahc-link --thread-pool-size=4 --input-hs perf-test/Setup.hs"
real    0m48,266s
user    4m2,818s
sys     0m2,243s

$ time stack build --exec "ahc-link --thread-pool-size=8 --input-hs perf-test/Setup.hs"
real    0m48,725s
user    4m8,747s
sys     0m3,086s
```

TerrorJack:
@gkaracha Thanks for the update! Indeed, it seems that adding parallelization is pragmatic only after we further cut down memory usage of the single-threaded case.

Issue #621, opened 2020-04-27
Implement parallel `ahc-ld`/`ahc-link`


TerrorJack:
**Is your feature request related to a problem? Please describe.**
Recent performance improvements has resulted in over 50% improvement in wall clock time when linking large programs. For further improvements, we should introduce parallelism in `ahc-ld` and `ahc-link`. There are multiple chances of parallelism which is explained in the next section.

**Describe the solution you'd like**

#### Chances of parallelism

* When loading archives and object files in `ahc-ld`, we can parallelize the deserialization of each object file. All object files are converted to `ByteString`s first, either via direct reading or `ArchiveEntry`, then the deserialization can be performed in parallel.
* After the gc-sections pass is run, the shrinked `AsteriusModule` should be fully evaluated, and this can be done in parallel as well.
* In the `binaryen` backend, we can parallelize the marshaling of different data segments and functions. `binaryen` will transparently switch to a new allocator when it notices it's allocating an IR node on a different thread, so we should ensure each Haskell worker thread is pinned using `forkOn`.

#### Method of parallelism

We cannot introduce additional dependencies like `parallel`, `monad-par` or `scheduler` here, since we need to strictly control our dependency surface. So we need to roll our minimal parallelism framework first.

The need for nested parallelism can be avoided for our use cases. A simple parallel loop should be sufficient:

```haskell
parallelFor :: Monoid r => Int -> [a] -> (a -> IO r) -> IO r
```

PR #624, opened 2020-04-28
[WIP] NFData Instance for AsteriusModule

gkaracha:
* Add `NFData` instances for most datatypes in `Asterius.Types`, and
* Utilize the new `NFData` instances to fully evaluate the shrinked `AsteriusModule` produced by `Asterius.Resolve.linkStart`.

After this PR is merged, we should be able to do a little better, by parallelizing `AsteriusModule`s evaluation (see https://github.com/tweag/asterius/issues/621).

### DIFF
--- a/asterius/src/Asterius/NFData/TH.hs
+++ b/asterius/src/Asterius/NFData/TH.hs
@@ -1,0 +1,55 @@
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE ViewPatterns #-}
+
+module Asterius.NFData.TH
+  ( genNFData,
+  )
+where
+
+import Control.DeepSeq
+import Data.List (foldl1')
+import Language.Haskell.TH
+
+genNFData :: Name -> Q [Dec]
+genNFData ty = do
+  TyConI dec <- reify ty
+  case dec of
+    DataD [] ((== ty) -> True) [] Nothing cons _
+      | length cons <= 0xFF ->
+        pure
+          [ InstanceD
+              Nothing
+              []
+              (AppT (ConT ''NFData) (ConT ty))
+              [ FunD
+                  'rnf
+                  [ Clause
+                      [ ConP
+                          (dataConName con)
+                          (map VarP vars)
+                      ]
+                      ( NormalB $
+                          if null vars
+                            then VarE '()
+                            else
+                              foldl1'
+                                (\acc x -> AppE (AppE (VarE 'seq) acc) x)
+                                (map (AppE (VarE 'rnf) . VarE) vars)
+                      )
+                      []
+                    | con <- cons,
+                      let vars = [mkName $ "a" <> show j | j <- [1 .. dataConFields con]]
+                  ]
+              ]
+          ]
+    _ -> fail $ "Asterius.NFData.TH.genNFData: " <> show dec
+
+dataConName :: Con -> Name
+dataConName (NormalC n _) = n
+dataConName (RecC n _) = n
+dataConName c = error $ "Asterius.NFData.TH.dataConName: " <> show c
+
+dataConFields :: Con -> Int
+dataConFields (NormalC _ fs) = length fs
+dataConFields (RecC _ fs) = length fs
+dataConFields c = error $ "Asterius.NFData.TH.dataConFields: " <> show c
--- a/asterius/src/Asterius/NFData/TH.hs
+++ b/asterius/src/Asterius/NFData/TH.hs
@@ -30,7 +30,7 @@ genNFData ty = do
                       ]
                       ( NormalB $
                           if null vars
-                            then VarE '()
+                            then ConE '()
                             else
                               foldl1'
                                 (\acc x -> AppE (AppE (VarE 'seq) acc) x)
--- a/asterius/src/Asterius/NFData/TH.hs
+++ b/asterius/src/Asterius/NFData/TH.hs
@@ -14,34 +14,33 @@ genNFData :: Name -> Q [Dec]
 genNFData ty = do
   TyConI dec <- reify ty
   case dec of
-    DataD [] ((== ty) -> True) [] Nothing cons _
-      | length cons <= 0xFF ->
-        pure
-          [ InstanceD
-              Nothing
-              []
-              (AppT (ConT ''NFData) (ConT ty))
-              [ FunD
-                  'rnf
-                  [ Clause
-                      [ ConP
-                          (dataConName con)
-                          (map VarP vars)
-                      ]
-                      ( NormalB $
-                          if null vars
-                            then ConE '()
-                            else
-                              foldl1'
-                                (\acc x -> AppE (AppE (VarE 'seq) acc) x)
-                                (map (AppE (VarE 'rnf) . VarE) vars)
-                      )
-                      []
-                    | con <- cons,
-                      let vars = [mkName $ "a" <> show j | j <- [1 .. dataConFields con]]
-                  ]
-              ]
-          ]
+    DataD [] ((== ty) -> True) [] Nothing cons _ ->
+      pure
+        [ InstanceD
+            Nothing
+            []
+            (AppT (ConT ''NFData) (ConT ty))
+            [ FunD
+                'rnf
+                [ Clause
+                    [ ConP
+                        (dataConName con)
+                        (map VarP vars)
+                    ]
+                    ( NormalB $
+                        if null vars
+                          then ConE '()
+                          else
+                            foldl1'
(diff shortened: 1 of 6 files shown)


gkaracha:
Currently ticks seem to go slightly down:
`diff master/ahc-ld.prof HEAD/ahc-ld.prof` gives
```diff
<       total time  =        9.94 secs   (9943 ticks @ 1000 us, 1 processor)
<       total alloc = 111,163,851,456 bytes  (excludes profiling overheads)
---
>       total time  =        9.39 secs   (9389 ticks @ 1000 us, 1 processor)
>       total alloc = 111,165,996,664 bytes  (excludes profiling overheads)
```
`diff master/ahc-link.prof HEAD/ahc-link.prof` gives
```diff
<       total time  =       45.82 secs   (45824 ticks @ 1000 us, 1 processor)
<       total alloc = 3,321,046,968 bytes  (excludes profiling overheads)
---
>       total time  =       45.49 secs   (45491 ticks @ 1000 us, 1 processor)
>       total alloc = 3,321,054,200 bytes  (excludes profiling overheads)
```
`diff master/ahc.prof HEAD/ahc.prof` gives
```diff
<       total time  =        0.28 secs   (282 ticks @ 1000 us, 1 processor)
<       total alloc = 132,290,048 bytes  (excludes profiling overheads)
---
>       total time  =        0.25 secs   (249 ticks @ 1000 us, 1 processor)
>       total alloc = 132,291,976 bytes  (excludes profiling overheads)
```

asterius/src/Asterius/NFData/TH.hs
```diff
@@ -0,0 +1,55 @@
...
+import Control.DeepSeq
+import Data.List (foldl1')
+import Language.Haskell.TH
+
+genNFData :: Name -> Q [Dec]
+genNFData ty = do
+  TyConI dec <- reify ty
+  case dec of
+    DataD [] ((== ty) -> True) [] Nothing cons _
+      | length cons <= 0xFF ->
```
TerrorJack, line 18: This is unnecessary. The reason it's there for the `Binary` TH instance generator is: we use a single byte for encoding data constructor tags, so we need to ensure the total number of datacons is no more than 256. But here we can have as many datacons as we like.

asterius/src/Asterius/Types/EntitySymbol.hs
```diff
@@ -21,6 +22,9 @@ newtype EntitySymbol = EntitySymbol GHC.FastString
   deriving newtype (Eq, Ord, Show, IsString, Semigroup, Monoid, GHC.Binary, GHC.Uniquable)
   deriving stock (Data)
 
+instance NFData EntitySymbol where
+  rnf = rwhnf -- TODO: Not entirely sure about this.
```
TerrorJack, line 26: This works since all `FastString` fields are strict. The todo can be dropped.

asterius/src/Asterius/Types/SymbolMap.hs
```diff
@@ -75,6 +76,9 @@ newtype SymbolMap a = SymbolMap (IM.IntMap (EntitySymbol, a))
   deriving newtype (Eq, Semigroup, Monoid)
   deriving stock (Data)
 
+instance NFData a => NFData (SymbolMap a) where
+  rnf = rwhnf
```
TerrorJack, line 80: This is wrong. Forcing the `IntMap` to WHNF doesn't force its values. I think the TH instance generator should also works here even if `IntMap` data constructors are not exported. But in case TH doesn't work, the correct instance is folding over the `IntMap` and applying `rnf` to the values along the way.

asterius/src/Asterius/Types/SymbolSet.hs
```diff
@@ -67,6 +68,9 @@ instance Show SymbolSet where
     showParen (p > 10) $
       showString "fromList " . shows (toList s)
 
+instance NFData SymbolSet where
+  rnf = rwhnf
```
TerrorJack, line 72: Wrong for the same reason as in `SymbolMap`.

asterius/src/Asterius/Resolve.hs
```diff
@@ -108,18 +110,19 @@ linkStart ::
...
-      { staticsSymbolMap = ss_sym_map,
-        functionSymbolMap = func_sym_map,
-        infoTableSet = makeInfoTableSet merged_m ss_sym_map,
-        Asterius.Types.LinkReport.tableSlots = tbl_slots,
-        staticMBlocks = static_mbs,
-        sptEntries = sptMap merged_m,
-        bundledFFIMarshalState = bundled_ffi_state
-      }
-  )
+  rnf merged_m
```
TerrorJack, line 113: The correct way to apply `rnf` is at `merged_m0`, as soon as gc-sections pass is over.

asterius/src/Asterius/Types/SymbolMap.hs
gkaracha: Right, I was sloppy here. Since `IntMap` is already an instance of `NFData` can't we also do
```haskell
instance NFData a => NFData (SymbolMap a) where
  rnf (SymbolMap m) = rnf m
```
?
If the `NFData` instance for `IntMap` is correct, I'd assume that it already does the traversal you describe, right?

TerrorJack: I just checked, `IntMap` already has `NFData` instance. We should be reusing that.

gkaracha: Cool, I'll replace it with the two lines above then :slightly_smiling_face: (same for `SymbolSet`).

asterius/src/Asterius/Resolve.hs
gkaracha: Hmmm, given that everything is pure, I don't understand exactly what you mean by "as soon as [the] gc-sections pass is over". Do you mean based on the data dependencies? Something like this?
```diff
-   merged_m1
-     | debug = addMemoryTrap merged_m0
-     | otherwise = merged_m0
+   merged_m1
+     | rnf merged_m0 `seq` debug = addMemoryTrap merged_m0
+     | otherwise = merged_m0

```

gkaracha: (well, basically `merged_m0 ``deepseq`` debug`)

asterius/src/Asterius/Types/SymbolMap.hs
```diff
@@ -75,6 +76,9 @@ newtype SymbolMap a = SymbolMap (IM.IntMap (EntitySymbol, a))
   deriving newtype (Eq, Semigroup, Monoid)
   deriving stock (Data)
 
+instance NFData a => NFData (SymbolMap a) where
```
TerrorJack, line 79: IMO it would be nice to put it to `deriving newtype`. Same applies for `SymbolSet`. Not for extra performance, just that we can save a bit of boilerplate code.

asterius/src/Asterius/Resolve.hs
TerrorJack: Introduce a `!merged_m0_evaluated = force merged_m0` binding between `merged_m0` and `merged_m1`, and use this in `merged_m1`.

asterius/src/Asterius/Resolve.hs
```diff
@@ -125,7 +127,7 @@ linkStart debug gc_sections verbose_err store root_syms export_funcs =
       | gc_sections = gcSections verbose_err store root_syms export_funcs
       | otherwise = fromCachedModule store
     merged_m1
-      | debug = addMemoryTrap merged_m0
+      | merged_m0 `deepseq` debug = addMemoryTrap merged_m0
```
TerrorJack, line 130: It's better to introduce a standalone binding for the forced `merged_m0` as mentioned in the earlier comment. A standalone binding means it'll show up in the profile report just like `merged_m0`, so we can check whether the full evaluation is a bottleneck.

gkaracha: Ah, right. Should be fixed now :slightly_smiling_face: 

(commit 0a6bf6b was pushed by accident, I was just experimenting locally).

gkaracha:
Since we are monitoring performance more closely these days, here is a quick comparison:

* `ahc` (before: 282 ticks, after: 253 ticks)
* `ahc-ld` (before: 9943 ticks, after: 10307 ticks)
* `ahc-link` (before: 45824 ticks, after: 45389 ticks)

TerrorJack:
The small cost in `ahc-ld` is acceptable, since the profile report now contains the cost center of evaluating the post gc-sections module.

PR #632, opened 2020-05-03
Fix "float unrepresentable in integer range" runtime error due to incorrect unsigned conversion

### LINKED ISSUE #629: RuntimeError: float unrepresentable in integer range
**Describe the bug**
```
kitchen-sink-exe.js:52 Uncaught (in promise) JSException "RuntimeError: float unrepresentable in integer range\n    at wasm-function[996]:0x340a1\n    at wasm-function[155]:0xbf31\n    at wasm-function[5697]:0x160d66\n    at wasm-function[5698]:0x160d97\n    at s.tick (http://localhost:4001/kitchen-sink-exe.js:24:4089)\n    at http://localhost:4001/kitchen-sink-exe.js:24:4551\n    at http://localhost:4001/kitchen-sink-exe.js:69:2729\n    at m (http://localhost:4001/kitchen-sink-exe.js:69:2839)\n    at i (http://localhost:4001/kitchen-sink-exe.js:69:2202)"
```

**To Reproduce**
Unfortunately it is very hard for me to trigger this in a reproduceable way - for small projects, I cannot get it to trigger. The large project on which it is triggering is here: https://github.com/meeshkan/plzwrk/tree/master/kitchen-sink . It can be compiled with `ahc-cabal` and then `ahc-dist`. The bug only comes up when one clicks on a button on the page.

**Expected behavior**
Expected not to crash.

**Environment**
 - Windows, but compiled using the provided docker image.

**Additional context**

I have tried to track down this bug by finding places in my code where floats are cast to integers, but I can't find anything that would result in this, so I'm thinking maybe something is going on with the internals of the wasm?

Thanks in advance for your help!

TerrorJack:
Closes #629. Thanks to @mikesol @InBetweenNames, we located a very old runtime bug. The symptom is a "float unrepresentable in integer range" runtime error, and the root cause is:

* We have quite a few rts builtin functions which are implemented in JavaScript, and imported into WebAssembly
* Since our word size is 64-bit and our WebAssembly functions mostly use `i64` in parameter/return types, the JavaScript functions need to be imported to use `f64`, and a layer of wrapper functions will generate casts between `i64` and `f64`.
* Our mechanism to generate the wrapper functions incorrectly used an unsigned cast for function results. When an rts function returns a negative result (e.g. `_hs_text_memcmp`), this will result in an error.

In the longer term we should be using the linear memory to pass FFI paramters/results in a uniform manner, without worrying about precision losses or signedness issues. But this PR suffices to fix the error for now.


InBetweenNames:
Happy to help!

PR #640, opened 2020-05-07
Add nofib testing

gkaracha:
Set up the nofib directory:
- [X] Gather nofib main programs and their arguments from the upstream nofib repo.
- [X] Flatten the structure (only two levels) and fix/remove symlinks.
- [X] Remove tests that don't apply to asterius yet and other redundant files.
- [ ] Replace local `Makefile`s with explicit `FAST_OPTS`, `NORM_OPTS` and `SLOW_OPTS` input files.

After that move on to CI-related tasks:
- [ ] Set up the native/asterius comparison and set it up as a standalone CI job.
- [ ] Generate V8 CPU/heap profiler report and serve them as CI artifacts, so we can examine the bottlenecks visually.


TerrorJack:
Additional notes:

* We'll likely need to evaluate the influence of different link-time options (e.g. enabling tail calls, yolo mode, different gc thresholds), since they can have a significant impact on the overall clock time. Also, certain `node` V8 command-line options (use `node --v8-options` to get a complete list; will discuss which options to test later)
* For getting a flame graph from V8 profiling reports, I tried a few tools and [`0x`](https://github.com/davidmarkclements/0x) seems to be pleasant to use.

PR #641, opened 2020-05-08
Adds plzwrk to the list of packages shipped in the Docker container

mikesol:


TerrorJack:
Thanks, but I think we'll hold this PR at the moment, given `plzwrk` is still in early days. Also, the `cabal.config`/`lts.sh` are harder to update when non-stackage packages are added.

PR #645, opened 2020-05-13
Fix `getArgs` support for node

### LINKED ISSUE #634: Implement support for `getArgs`
We are about to add `nofib` support to examine runtime performance quantitatively. The `nofib` programs usually read some arguments with `getArgs` to decide how long should the tests run. However, we don't support `getArgs` in both node/browser for now.

* [ ] Implement `getArgs` support. Whatever `getArgs` returns should be a part of the JavaScript `.req.mjs` script.
* [ ] Implement a linker flag to specify how `getArgs` should get its input; empty, or `process.argv` when available, or a literal specified via the linker flag.

TerrorJack:
Closes #634.

This PR fixes `getArgs` for `node`. Say we compile `test.hs` and run `node test.mjs extra flags`, `getArgs` will properly return `["extra", "flags"]`.

The current implementation is just enough to support #640. Limitations compared to native `getArgs`:

* No support for `setProgArgv`.
* Total size of arguments must not exceed 1024 bytes.
* No handling for `+RTS ...` flags. See https://gitlab.haskell.org/ghc/ghc/-/blob/ghc-8.8/rts/RtsFlags.c#L612 for complete logic to handle them, which IMO is an unnecessary complexity to add for now.
* No support for `getFullArgs`.

### DIFF
--- a/asterius/rts/node/default.mjs
+++ b/asterius/rts/node/default.mjs
@@ -13,18 +13,29 @@ class Posix {
     this.errno = 0;
     Object.seal(this);
   }
-  getProgArgv(argc, argv) {
+  getProgArgv(argc, argv_buf) {
     const e = new TextEncoder(),
       arg_bufs = process.argv.slice(2).map((s) => e.encode(s)),
+      argv_header_size = (1 + arg_bufs.length) * 8,
       argv_total_size =
-        (1 + arg_bufs.length) * 8 +
+        argv_header_size +
         arg_bufs.reduce((acc, buf) => acc + buf.byteLength + 1, 0);
     if (argv_total_size > 1024) {
       throw new WebAssembly.RuntimeError(
         `getProgArgv: exceeding buffer size for ${process.argv}`
       );
     }
-    this.memory.i64Store(argc, 1+arg_bufs.length);
+    this.memory.i64Store(argc, 1 + arg_bufs.length);
+    let p0 = argv_buf + 8,
+      p1 = argv_buf + argv_header_size;
+    arg_bufs.forEach(arg_buf => {
+      this.memory.i64Store(p0, p1);
+      p0 += 8;
+      this.memory.expose(p1, arg_buf.byteLength, Uint8Array).set(arg_buf);
+      p1 += arg_buf.byteLength;
+      this.memory.i8Store(p1, 0);
+      p1 += 1;
+    });
   }
   get_errno() {
     return this.errno;
--- a/asterius/rts/node/default.mjs
+++ b/asterius/rts/node/default.mjs
@@ -14,8 +14,17 @@ class Posix {
     Object.seal(this);
   }
   getProgArgv(argc, argv) {
-    this.memory.i64Store(argc, 0);
-    this.memory.i64Store(argv, 0);
+    const e = new TextEncoder(),
+      arg_bufs = process.argv.slice(2).map((s) => e.encode(s)),
+      argv_total_size =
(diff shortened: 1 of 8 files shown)


asterius/src/Asterius/Builtins/Env.hs
```diff
@@ -0,0 +1,49 @@
...
+            { paramTypes = [F64, F64],
+              returnTypes = []
+            }
+      }
+  ]
+
+envCBits :: AsteriusModule
+envCBits = envArgvBuf <> envGetProgArgv
+
+envArgvBuf :: AsteriusModule
```
gkaracha, line 31:
```suggestion
-- | Data segment used to store runtime arguments. In normal native programs,
the arguments passed to a program @argv@ exist in special memory pages and are
managed by the OS kernel. Since we cannot do this here, we instead reserve a
data segment of 1KB of memory to be used for this purpose (see @getProgArgv@ in
@rts/node/default.mjs@).
envArgvBuf :: AsteriusModule
```

asterius/rts/node/default.mjs
```diff
@@ -13,6 +13,30 @@ class Posix {
     this.errno = 0;
     Object.seal(this);
   }
+  getProgArgv(argc, argv_buf) {
+    const e = new TextEncoder(),
+      arg_bufs = process.argv.slice(2).map((s) => e.encode(s)),
+      argv_header_size = (1 + arg_bufs.length) * 8,
+      argv_total_size =
+        argv_header_size +
+        arg_bufs.reduce((acc, buf) => acc + buf.byteLength + 1, 0);
```
gkaracha, line 22:
```suggestion
        // All strings are \0-terminated, hence the +1
        arg_bufs.reduce((acc, buf) => acc + buf.byteLength + 1, 0); 
```

asterius/rts/node/default.mjs
```diff
@@ -13,6 +13,30 @@ class Posix {
...
+    if (argv_total_size > 1024) {
```
gkaracha, line 23:
```suggestion
    // The total size (in bytes) of the runtime arguments cannot exceed the 1KB
    // size of the data segment we have reserved. If you wish to change this
    // number, you should also update envArgvBuf in Asterius.Builtins.Env.
    if (argv_total_size > 1024) {
```

asterius/src/Asterius/Builtins/Env.hs
TerrorJack: Missing `--` prefix for rest of the lines.

gkaracha: :facepalm: so true. I just updated the suggestion.

TerrorJack:
@gkaracha Go ahead with the test then; remember to pull the weeds in the code snippet you pasted.

gkaracha:
> @gkaracha Go ahead with the test then; remember to pull the weeds in the code snippet you pasted.

https://github.com/tweag/asterius/pull/645/commits/154673550467858719e46c5cb5269623102a825d should do it :slightly_smiling_face:

asterius/test/argv/argv.hs
```diff
@@ -0,0 +1,11 @@
+import Asterius.Types
+import Data.Coerce
+import System.Environment
+
+foreign import javascript "console.log($1)" js_print :: JSVal -> IO ()
+
+js_print_string :: String -> IO ()
```
TerrorJack, line 7: Don't forget to remove the unnecessary imports and definitions.

Issue #345, opened 2019-11-26
ahc-boot No such file @/var/.../ar29968-0.rsp


BebeSparkelSparkel:
Trying to run this on macOS 10.15.1 but it is inserting `@` at the beginning of the file path

```shell
> ahc-boot
ar: @/var/folders/zj/p9mj9djj3c3b4vjw5yzq1ckr0000gn/T/ar29968-0.rsp: No such file or directory
callProcess: ar "-r" "-c" "/Users/williamrusnack/Documents/asterius/.stack-work/install/x86_64-osx-custom-asterius/5712d8350648752475833347e3e5d3b317375b52376aecf5a470fc0d935bd03d/8.6.5/share/x86_64-osx-ghc-8.6.5/asterius-0.0.1/.boot/asterius_lib/rts/libHSrts.a" "@/var/folders/zj/p9mj9djj3c3b4vjw5yzq1ckr0000gn/T/ar29968-0.rsp" (exit 1): failed
```

BebeSparkelSparkel:
[Asterius.Boot L145](https://github.com/tweag/asterius/blob/7778e06e0a3c886f0f19835a156d6f87374acc23/asterius/src/Asterius/Boot.hs#L145)
adds an `@` that makes `ar` fail.

TerrorJack:
Thanks for reporting. I have a WIP branch for fixing & testing darwin support [here](https://github.com/tweag/asterius/tree/wip-darwin-ci); after fixing the `@` issue of booting, it now [fails](https://circleci.com/gh/tweag/asterius/7640) with the following error message when running `configure` of an autoconf package:

```
Configuring ghc-prim-0.5.3...
configure: WARNING: unrecognized options: --with-compiler
checking for gcc... /usr/bin/gcc
checking whether the C compiler works... no
configure: error: in `/Users/distiller/project/.stack-work/install/x86_64-osx-custom-asterius/1d778aa5cc4903dd0f7b6344c055c493b33c2d1b0672db9d92e829c6c4299605/8.6.5/share/x86_64-osx-ghc-8.6.5/asterius-0.0.1/.boot/dist/dist/ghc-prim/build':
configure: error: C compiler cannot create executables
See `config.log' for more details
ahc-boot: user error (boot failure)
```

I'm not a darwin toolchain expert, so I'm stuck on this step. Would be great if you or someone else provide a bit of guidance here.

BebeSparkelSparkel:
[...]

PR #663, opened 2020-05-22
WIP Implement ahc-ar

### LINKED ISSUE #649: Implement `ahc-ar`
When using `ahc-cabal` to compile library components, it calls `ar` to make the static library archives from the object files. Later at link time, `ahc-ld` would decode the objects in the archives before further linking logic.

Relying on system `ar` introduces a hidden dependency on gnu `ar`; the default `ar` on `darwin` doesn't support the `.rsp` files passed to it, which results in boot failures, as seen in #345. It's possible to workaround the issue by installing gnu `binutils` on `darwin`, but that's a heavyweight change in the user environment we'd like to avoid.

The more ideal solution is implementing a simple `ahc-ar` program and specify it as the `ar` program in the `ahc-cabal` default configuration flags. There are additional benefits besides fixing `darwin` support, see below.

* [ ] Implement `ahc-ar` which recombines multiple object files into a single one, and output the combined object file as the "archive" file. Cabal doesn't care whether `.a` is a real archive and it's going to be handled by `ahc-ld` anyway. This will increase some compile time in exchange for shorter link time.
* [ ] Simplify `ahc-ld` and TH loader logic, handle `.a` files just like regular `.o` files
* [ ] Make `ahc-cabal` use `ahc-ar` as the default `ar` program

gkaracha:
Closes #649.

### DIFF
--- a/asterius/app/ahc-ar.hs
+++ b/asterius/app/ahc-ar.hs
@@ -1,8 +1,98 @@
+{-# LANGUAGE OverloadedStrings #-}
+
 import System.Environment.Blank
 import System.Process (callProcess)
+import qualified Data.ByteString as BS
+import System.IO (IOMode(..), withFile, Handle, hFileSize, hFlush)
+import Control.Monad (forM_, when)
+import qualified Data.ByteString.Char8 as BSC
+-- import System.Directory (doesFileExist)
+
+-- TODOs:
+-- * Add proper checks (doesFileExist, etc.)
+-- * Add proper flag parsing
+-- * Maybe work on a temporary file first and copy the results afterwards?
+-- * Cleanup and ormolize
+-- * Still unclear: should we use withBinaryFile etc. instead?
+
+-- Some resources:
+--   https://github.com/haskell/cabal/blob/master/Cabal/Distribution/Simple/Program/Ar.hs
+--   https://en.wikipedia.org/wiki/Ar_(Unix)
+--   https://pubs.opengroup.org/onlinepubs/007908799/xcu/ar.html
+--   https://www.freebsd.org/cgi/man.cgi?query=ar&sektion=5&manpath=4.3BSD+NET%2F2
 
 main :: IO ()
 main = do
   args <- getArgs
   callProcess "ar" args -- For now just call GNU ar
+
+-- | Get all arguments, looking through file arguments.
+--
+-- TODO: Adjust this function to be faithful to the behavior described in @man ar@:
+--
+-- Read command-line options from file. The options read are inserted in place of
+-- the original @file option. If file does not exist, or cannot be read, then the
+-- option will be treated literally, and not removed.
+--
+-- Options in file are separated by whitespace. A whitespace character may be
+-- included in an option by surrounding the entire option in either single or
+-- double quotes. Any character (including a backslash) may be included by
+-- prefixing the character to be included with a backslash. The file may itself
+-- contain additional @file options; any such options will be processed
+-- recursively.
+getArgsRecursively :: IO [String]
+getArgsRecursively = getArgs >>= expandAtOptions
+  where
+    expandAtOptions :: [String] -> IO [String]
+    expandAtOptions opts = concat <$> mapM expandAtOption opts
+
+    expandAtOption :: String -> IO [String]
+    expandAtOption ('@':path) = readFile path >>= expandAtOptions . words
+    expandAtOption opt = return [opt]
+
+-----------------------------------------------------------------------------
+
+-- | Create a library archive from a bunch of object files.
+createArchive :: Bool -> FilePath -> [FilePath] -> IO ()
+createArchive is_truncation_allowed arFile objFiles =
+  withFile arFile WriteMode $ \ah -> do
+    BS.hPut ah "!<arch>\x0a" -- ASCII Magic String (8 bytes)
+    forM_ objFiles $ hPutObjFile ah
+  where
+    hPutObjFile :: Handle -> FilePath -> IO ()
+    hPutObjFile ah filename = do
+      withFile filename ReadMode $ \oh -> do
+        let fileID = mkFileID filename
+        fileSize <- hFileSize oh
+        let bsFileSize = BSC.pack $ take 10 $ show fileSize ++ repeat ' '
+        hPutFileHeader ah fileID bsFileSize
+        hCopyContents fileSize oh ah
+        hFlush ah -- Better safe than sorry
+
+    -- TODO: assert that filename and filesize have the proper size.
+    hPutFileHeader :: Handle -> BS.ByteString -> BS.ByteString -> IO ()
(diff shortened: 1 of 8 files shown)


gkaracha:
@TerrorJack This is still a rough first draft but maybe it's time to have a look, tell me if I am going the right way about it? It seems to be working just fine on my Ubuntu, but I have no idea if it would really work on darwin (e.g. fix issue https://github.com/tweag/asterius/issues/345).

gkaracha:
Nah, scrap that. I missed an error from [getGNUArchEntries](https://hackage.haskell.org/package/ghc-8.6.1/docs/src/Ar.html#getGNUArchEntries) due to name truncation. I'll fix that first and we can discuss options later.

ProofOfKeags:
I can test on Darwin if you want to fish for what issues you might run into.

gkaracha:
> I can test on Darwin if you want to fish for what issues you might run into.

Ah, that'd be great! I will ping when the patch is ready for testing.

gkaracha:
OK, @TerrorJack you have a look now. We still have to decide how much we want to deviate from what GNU `ar` does. The patch currently
* just ignores all arguments but the first `.a` file and the list of `.o` files. Shall we warn or do something with the rest?
* does not parse quoted arguments correctly (e.g. `"this is a filename with spaces in it.o"`), it uses `words` to space-separate the arguments. Do we need that?
* parses `@file` arguments recursively (as specified by `man ar`). Do we need that behavior?

TerrorJack:
@gkaracha

* No need to warn, since `ahc-ar` is only called by `Cabal` and we know exactly how it is called
* Use `unlines` instead of `unwords`, see https://haskell-code-explorer.mfix.io/package/Cabal-2.4.1.0/show/Distribution/Simple/Program/ResponseFile.hs#L34 for how the response file is generated.
* As discussed earlier, the recursive parsing behavior should be removed for simplicity. There's no recursion in `Cabal`-generated response files.

TerrorJack:
One place we forgot to modify: the `boot.sh` boot script compiles libraries in two ways: either calling `ahc-cabal`, or calling a `Setup` program and completing the `configure`/`build`/`install` steps manually. This PR only affects `ahc-cabal` so far; we also need to modify `Asterius.Boot` and add `--with-ar=<path to ahc-ar>` to the default configure options.

gkaracha:
@TerrorJack

> see https://haskell-code-explorer.mfix.io/package/Cabal-2.4.1.0/show/Distribution/Simple/Program/ResponseFile.hs#L34 for how the response file is generated.

Nice, thanks for the source. This means that we have to undo the effects of [`escapeResponseFileArg`](https://haskell-code-explorer.mfix.io/package/Cabal-2.4.1.0/show/Distribution/Simple/Program/ResponseFile.hs#L49) too, right?

TerrorJack:
I don't think there's need to unescape object file paths for now. The bizzare object file names have never occured in the field. If later we encounter a crash in third party libraries we can fix this up.

TerrorJack:
@gkaracha Please cherry pick https://github.com/tweag/asterius/commit/8bd5b82010f19181bf6ab8011c212abc43c28bda on this branch and re-test. For unknown reasons, `--with-ar=` must be hard-wired in `boot.sh`. If you add `-v3` to configure options you can show that the system `ar` is still used previously.

gkaracha:
> @gkaracha Please cherry pick [8bd5b82](https://github.com/tweag/asterius/commit/8bd5b82010f19181bf6ab8011c212abc43c28bda) on this branch and re-test.

Done: https://buildkite.com/tweag-1/asterius/builds/609. Fingers crossed.

TerrorJack:
@gkaracha And also https://github.com/tweag/asterius/commit/0085dcf4c7b3200362092efbf8c7cdc751b5d3db. The patch just gives up the reparsing/combining logic of object files.

The reason for the patch is a segmentation fault in macOS booting as shown in CI. I suspect there might have been an OOM condition, and giving up object recombining seems to make `ahc-ar` work on macOS. (Although it's failing for something else now)

TerrorJack:
I apologize for the "good first issue" tag in #649 btw. I just removed it. :facepalm:

gkaracha:
> @gkaracha And also [0085dcf](https://github.com/tweag/asterius/commit/0085dcf4c7b3200362092efbf8c7cdc751b5d3db). The patch just gives up the reparsing/combining logic of object files.

Yes, I think that's a good idea in general, it is a change in logic we best include in a separate PR.

> The reason for the patch is a segmentation fault in macOS booting as shown in CI. I suspect there might have been an OOM condition, and giving up object recombining seems to make `ahc-ar` work on macOS. (Although it's failing for something else now)

Hmmmm, something similar is blocking https://github.com/tweag/asterius/pull/622 (it also OOM for 2 or more cores). If/When we find the source of this I will revisit https://github.com/tweag/asterius/pull/622 to see if it's unblocked.

> I apologize for the "good first issue" tag in #649 btw. I just removed it.

No worries about that! (but still, thanks for removing :stuck_out_tongue:)

asterius/app/ahc-ar.hs
```diff
@@ -0,0 +1,78 @@
...
+-- Copyright   :  (c) 2018 EURL Tweag
+-- License     :  All rights reserved (see LICENCE file in the distribution).
+--
+-- This module contains the implementation of @ahc-ar@, a system-agnostic,
+-- partial implementation of GNU @ar@ for creating archive files from a set of
+-- object files. The shape of the contents of the generated archive files is as
+-- explained [here](https://en.wikipedia.org/wiki/Ar_(Unix)#File_header). For
+-- all fields (except the file size), we use default values, as follows:
+--
+-- > ---------+-----------------------------+----------------+
```
TerrorJack, line 12: Nice diagram here, but the module header documentation is now out-of-date I think, so still better remove it (or change it to mention we use `Ar` in GHC API to generate gnu ar files)

TerrorJack:
@ProofOfKeags This PR solves the `ar` problem on macOS, but not sufficient to fully fix macOS support; there's another error as in https://app.circleci.com/pipelines/github/tweag/asterius/1697/workflows/97316b3c-006c-4e78-9e40-c8e14560563f/jobs/13829.

asterius/app/ahc-ar.hs
gkaracha: revamped the comments: https://github.com/tweag/asterius/pull/663/commits/73643c6e0d50780a846ff10b5bb0d33c7e2b731f

ProofOfKeags:
@TerrorJack Do you know the cause of the new error? or does that need to be investigated?

TerrorJack:
@ProofOfKeags See https://github.com/tweag/asterius/issues/674#issuecomment-641789171. It'll take some time to figure out the solution; right now a possible workaround is adding gnu gcc to replace the xcode stuff in `PATH` in the shell which you call asterius, but I don't have time to test whether it'll actually work.

Issue #665, opened 2020-05-27
Avoid loading all object files at link-time


TerrorJack:
**Is your feature request related to a problem? Please describe.**
We've been bothered by the excessive memory usage of the linker for a long time. There has been various tricks (e.g. lazy serialization of `Map` values) and efforts (e.g. #605) to improve it. In this ticket, we seek a deeper refactoring that addresses a technical debt: needing to load all objects at link-time.

Until now, the linker codebase mostly works with the `AsteriusModule` type which consists of several symbol-indexed immutable hash maps. This makes implementation simple, but retains a lot of heap data. This is especially true for the input to the gc-sections pass, since most input entities will be thrown away, but they had to be present in the memory for some time.

In the past, we used laziness to reduce memory consumption here. The hash maps in `AsteriusModule` has lazy values; when deserializing, each lazy value is a thunk which will decode to the actual entity, and the thunks retain the reference to the whole `ByteString` which is the object file content. This will take a lot less memory and time than strictly deserializing and having the decoded entity in the heap. Still, we have to perform this kind of lazy-loading for all object files, and this requires at least as much memory as the file sizes of input archives and objects.

Production linkers like `lld` don't load all object files at link-time. It's about the time we do the same and replace the previous "lazy loading" trick with a more decent solution.

**Describe the solution you'd like**

* Refactor the `AsteriusCachedModule` type and serialization logic. The new logic should support:
  * Only reading the symbol dependency map, without touching the rest of the file
  * Given an entity's symbol, deserialize it starting from a certain offset of the file, without reading the whole file.
* The gc sections pass is split into smaller passes:
  * Given the input (archives, objects, in-memory `AsteriusModule`), scan the files and build up the symbol dependency map. After this completes, there shouldn't be heap resident blobs of input file contents.
  * Run gc sections using the symbol dependency map
  * We now have a list of symbols, and also the origins of these symbols (which object file and what offset into that file). Deserialize the entities indexed by those symbols and form the `AsteriusModule` as the gc section output. Before we introduce proper streaming logic into the rest of linker logic, the entities should still be lazily decoded; each thunk will read its own object file and perform decoding on its offset.

This should further reduce linker memory usage by avoiding retaining object file content blobs in memory.

**Describe alternatives you've considered**
[...]

PR #666, opened 2020-05-28
WIP Revamp linker loading (do not load all object files during linking)

gkaracha:
Closes https://github.com/tweag/asterius/issues/665. This PR is a WIP.

Roadmap for addressing **the first bullet** of https://github.com/tweag/asterius/issues/665:

* [X] Define new type for `dependencyMap`s, and serialization/deserialization logic for it. 
* [ ] Add an index in `AsteriusCachedModule` after the `dependencyMap` field, as a map from `EntitySymbol` to both the file that the entity originates from and the offset in the file. Though initially the offset is enough, when we concatenate modules we end up having entities from different files so we need both the file and the offset in the file to pinpoint the location of the entity.
* [ ] Implement retrieval of entities without reading any other part of the file.

Roadmap for addressing **the second bullet** of https://github.com/tweag/asterius/issues/665:
* [ ] TBD


TerrorJack:
In #665, the "dependency map" is just an abstract concept, it doesn't have to be a concrete data type. Adding a separate data type here (which is just another wrapper of `SymbolMap`) is a bit overwork IMO.

gkaracha:
You might be right, but I cannot be sure about how much we'll end up passing it around. If it is a lot, I'd say it deserves it's own type. Let's see how it goes and if it ends up being just noise I'll remove it before we merge this :slightly_smiling_face:

gkaracha:
@TerrorJack lest we forget:

Since we need (at least parts of) the `FFIMarshalState` to build the set of root symbols (see https://github.com/tweag/asterius/blob/master/asterius/src/Asterius/Passes/GCSections.hs#L43), having some parts of the `FFIMarshalState` in `AsteriusCachedModule` looks unavoidable.

gkaracha:
After discussing this with @TerrorJack, the current plan is to follow a different approach than the one originally discussed in https://github.com/tweag/asterius/issues/665. We avoid the hassle of adding an index inside the object files and instead:
* Make object files contain (a) the `dependencyMap` and (b) the actual `AsteriusModule`. Before `gcSections` we only deserialize the dependency map, and after `gcSections` we deserialize the asterius module. This means that both should be length-prefixed so that we can skip them.
* After `gcSections` has determined which symbols we need to keep, we do indeed read all object and archive files that we need to read, but on-the-fly keep the parts that we need. Probably this means some extra work for the garbage collector, but it still avoids keeping all the data in memory.

PR #678, opened 2020-06-12
Implement full-blown Asterius.Ar based on GHC's Ar

gkaracha:
Though we have already moved away from GNU `ar` (https://github.com/tweag/asterius/pull/663), we still use [GHC's `loadAr`](https://haskell-code-explorer.mfix.io/package/ghc-8.6.5/show/main/Ar.hs#L240). We plan to make many more changes to our treatment of archive files (https://github.com/tweag/asterius/pull/666), so in this PR we roll out our custom implementation of `Ar` based on that of GHC.

**Edit:** Sidenote: Since `ahc-cabal` internally uses `cabal`, and `cabal` internally traverses archive files and edits file metadata (see [Cabal.Distribution.Simple.Program.Ar.wipeMetadata](https://github.com/haskell/cabal/blob/b256029ac51d0c837bb558d7e3a6eb54491dc44a/Cabal/Distribution/Simple/Program/Ar.hs#L125-L191)), the general shape of archive files must remain the same, so our implementation is still based on `Archive` and `ArchiveEntry`, as defined in GHC's Ar.


gkaracha:
Changed the name and the description of the PR to better reflect what it does. Our plans for customizing the treatment of archive files will be implemented by future PRs.

Original description:
> At the moment we rely on [GHC's `loadAr`](https://haskell-code-explorer.mfix.io/package/ghc-8.6.5/show/main/Ar.hs#L240) which loads a whole archive file from disk. Obviously this does not play well with the incremental approach we hope to achieve with PR https://github.com/tweag/asterius/pull/666. This PR aims to rectify this. We do the following:
>
> * Replace calls to GHC's `loadAr` with more fine-grained variants.
> * Restrict the shape of archive files produced by `ahc-ar` and consumed by `loadAr` to look like GNU-style archives only.
> * Add an index entry to the archive files produced by `ahc-ar` and consumed by `loadAr`. This entry contains the dependency map derived from combining all dependency maps from the contained object files. Hence, before `gcSections` we should be able to get the dependency map of an archive file without having to read the whole thing.
>
> This is work in progress.

PR #700, opened 2020-06-29
Switch to new generation `inline-js-core`

TerrorJack:
This PR switches from `develop` branch of `inline-js-core` to the `v2` new generation branch, which will soon become the new `master`. The new generation `inline-js-core` offers many benefits over the previous implementations; details will follow in `inline-js` design doc shortly.


TerrorJack:
Closed in favor of #701.

PR #702, opened 2020-07-03
Prefix object files with a "magic" string for faster failure

gkaracha:
When loading object files (in Asterius.Ar.loadArchive and Asterius.Ld.loadTheWorld) we silently ignore malformed object files, which likely come from gcc-compiled files in libraries. Adding a magic string at the beginning of each object file should speed things up a little, since we can detect such deserialization failures faster.

### DIFF
--- a/asterius/src/Asterius/Types.hs
+++ b/asterius/src/Asterius/Types.hs
@@ -3,6 +3,8 @@
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StrictData #-}
 {-# LANGUAGE TemplateHaskell #-}
@@ -59,9 +61,12 @@ import Asterius.Types.SymbolMap (SymbolMap)
 import qualified Asterius.Types.SymbolMap as SM
 import Asterius.Types.SymbolSet (SymbolSet)
 import qualified Asterius.Types.SymbolSet as SS
+import qualified Binary as GHC
 import Control.Exception
+import Control.Monad
 import qualified Data.ByteString as BS
 import Data.Data
+import Data.Foldable
 import qualified Data.Map.Lazy as LM
 import Foreign
 import qualified Type.Reflection as TR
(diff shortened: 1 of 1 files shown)


gkaracha:
Indeed, the `.prof` files show that the ticks go down a little:

ahc:
```diff
<       total time  =        0.20 secs   (205 ticks @ 1000 us, 1 processor)
<       total alloc = 139,174,112 bytes  (excludes profiling overheads)
---
>       total time  =        0.19 secs   (192 ticks @ 1000 us, 1 processor)
>       total alloc = 139,177,056 bytes  (excludes profiling overheads)
```
ahc-ld
```diff
<       total time  =       10.76 secs   (10758 ticks @ 1000 us, 1 processor)
<       total alloc = 104,578,957,128 bytes  (excludes profiling overheads)
---
>       total time  =       10.28 secs   (10283 ticks @ 1000 us, 1 processor)
>       total alloc = 104,579,440,600 bytes  (excludes profiling overheads)
```
ahc-link
```diff
<       total time  =       77.74 secs   (77738 ticks @ 1000 us, 1 processor)
<       total alloc = 3,244,640,592 bytes  (excludes profiling overheads)
---
>       total time  =       75.09 secs   (75088 ticks @ 1000 us, 1 processor)
>       total alloc = 3,244,682,064 bytes  (excludes profiling overheads)
```

gkaracha:
@TerrorJack, since #666 was getting massive, I thought I'd take some parts of it that are useful regardless and make separate PRs from them. I think this one is a good candidate.

asterius/src/Asterius/Types.hs
```diff
@@ -140,6 +145,30 @@ instance Semigroup AsteriusCachedModule where
...
+  put_ bh AsteriusCachedModule {..} = do
+    putObjectMagic bh
+    GHC.put_ bh dependencyMap
+    GHC.put_ bh fromCachedModule
+
+objectMagic :: BS.ByteString
+objectMagic = "!<asterius>\n"
+
+putObjectMagic :: GHC.BinHandle -> IO ()
+putObjectMagic bh = for_ (BS.unpack objectMagic) (GHC.putByte bh)
```
TerrorJack, line 164: Using `putByte`/`getByte` is inefficient, we should use the `Binary` instance of `ByteString` and call `put_`/`get` directly.

gkaracha: Hmmm, the reason I avoided it is because the instance for `ByteString` uses a length-prefixed encoding, which makes it non constant-time. If it is a malformed object file, we could read any kind of length there. I'd use `peek`/`poke` (e.g. as for the [`Word` instances](https://haskell-code-explorer.mfix.io/package/ghc-8.6.5/show/utils/Binary.hs#L1275)) but this part of the interface is not exposed.

TerrorJack: In that case, we should still use `putPrim`/`getPrim`. For putting, use `Data.ByteString.Unsafe.unsafeUseAsCStringLen` to expose the internal pointer/length of `objectMagic` and then use `putPrim` to put it; for `getPrim`, after getting the pointer, no need to reconstruct a `ByteString` for comparing against `objectMagic`, since we can `memcmp` directly.

gkaracha: Neither `putPrim` nor `getPrim` is exported from `Binary` though :/

TerrorJack: Just noticed that. Yeah, then the current implementation is fine as it is.

gkaracha:
Just a note: if it ends up being problematic performance-wise in the future, we could always use a `Word64` magic which should be constant-time and efficiently implemented.

PR #703, opened 2020-07-03
Add missing Functor/Foldable/Traversable instances for SymbolMap

gkaracha:


gkaracha:
This one is also outsourced from #666.

PR #706, opened 2020-07-07
[WIP] Use Word64 magic (instead of ByteString) for efficiency

gkaracha:


gkaracha:
Quick comparison of d14dfcf and https://github.com/tweag/asterius/commit/189053a02ad660dc24866bf0312fcbad230ca83e gives

ahc
```diff
<       total time  =        0.18 secs   (179 ticks @ 1000 us, 1 processor)
<       total alloc = 139,174,872 bytes  (excludes profiling overheads)
---
>       total time  =        0.18 secs   (182 ticks @ 1000 us, 1 processor)
>       total alloc = 139,178,656 bytes  (excludes profiling overheads)
```
ahc-ld
```diff
<       total time  =       10.62 secs   (10616 ticks @ 1000 us, 1 processor)
<       total alloc = 104,579,907,256 bytes  (excludes profiling overheads)
---
>       total time  =       10.47 secs   (10472 ticks @ 1000 us, 1 processor)
>       total alloc = 104,581,636,440 bytes  (excludes profiling overheads)
```
ahc-link
```diff
<       total time  =       70.73 secs   (70731 ticks @ 1000 us, 1 processor)
<       total alloc = 3,244,651,672 bytes  (excludes profiling overheads)
---
>       total time  =       73.20 secs   (73197 ticks @ 1000 us, 1 processor)
>       total alloc = 3,244,622,112 bytes  (excludes profiling overheads)
```
I have to admit I'm a little perplexed; I expected using a `Word64` to be in all ways superior to using a 12-byte long `ByteString` (PR https://github.com/tweag/asterius/pull/702). Must be missing something.

TerrorJack:
Note that the profile job is only run once and there may be a certain degree of random fluctuation, so the pipeline better be run multiple times to collect the mean tick number for more certainty.

Also, `Word64`'s `Binary` instance isn't terribly efficient, since it really performs 8 `Word8` operations for endian-independence. So if you'd like to squeeze out a bit of extra performance here, you can seek to write your own `put`/`get` for `Word64` which operates on the buffer directly..

gkaracha:
> Note that the profile job is only run once and there may be a certain degree of random fluctuation, so the pipeline better be run multiple times to collect the mean tick number for more certainty.

:+1: thanks for the tip!

> Also, `Word64`'s `Binary` instance isn't terribly efficient, since it really performs 8 `Word8` operations for endian-independence. So if you'd like to squeeze out a bit of extra performance here, you can seek to write your own `put`/`get` for `Word64` which operates on the buffer directly..

Ah, right, I thought that the `Word64` instance was a little faster because it uses `getPrim h 8` but then it peeks each byte separately. I'll see if I can leverage `withBinBuffer` to make this a little faster.

gkaracha:
Given the lack of access to `Binary`'s underlying buffer I doubt that we'll get any significant improvement by switching to `Word64`; we might as well stay with what we have. I'm closing this.

PR #707, opened 2020-07-07
Refactor gcSections to resolve symbols first and gc after

gkaracha:
Being able to partition `gcSections` into two separate phases without problems (all tests must pass and there should be no performance regressions) is a great step in the right direction. If this PR works as expected, we shall edit the second phase (module filtering/creation from disk) afterwards in a separate PR, instead of doing all at once (PR #666).

### DIFF
--- a/asterius/src/Asterius/Passes/GCSections.hs
+++ b/asterius/src/Asterius/Passes/GCSections.hs
@@ -76,7 +76,7 @@ gcModule mod_syms export_funcs m =
   where
     statics = staticsMap m `SM.restrictKeys` mod_syms
     functions = functionMap m `SM.restrictKeys` mod_syms
-    spt_map = sptMap m `SM.restrictKeys` SM.keysSet statics
+    spt_map = sptMap m `SM.restrictKeys` mod_syms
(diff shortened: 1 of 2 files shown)


asterius/src/Asterius/Passes/GCSections.hs
```diff
@@ -3,92 +3,99 @@
...
+      ffiMarshalState =
+        FFIMarshalState
+          { ffiImportDecls = ffi_imports,
+            ffiExportDecls = ffi_exports
+          }
+    }
+  where
+    statics = staticsMap m `SM.restrictKeys` mod_syms
+    functions = functionMap m `SM.restrictKeys` mod_syms
+    spt_map = sptMap m `SM.restrictKeys` SM.keysSet statics
```
TerrorJack, line 79: Indeed, there is guarantee that all SPT entries have data segment symbol keys, so we can use `SM.keysSet statics`. But `mod_syms` should also work and we can use that to avoid computing `SM.keysSet statics`.

gkaracha: Thanks!

PR #714, opened 2020-07-09
Delete law-violating obsolete Semigroup/Monoid instances for LinkReport

gkaracha:
~~Though lawful, these~~ These instances (the `Semigroup` one in particular) assume that `tableSlots` and `staticMBlocks` will be set correctly externally, which is weird and could easily get them out-of-sync. Fortunately, they don't seem to be used anywhere (the only place that seems to use them in `linkStart` sets all the fields explicitly anyway).


gkaracha:
Actually, they are not even lawful now that I think about it: `r <> mempty /= r` if `tableSlots r /= 0`.

PR #719, opened 2020-07-22
[WIP] Use BigInt for I64 FFI

### LINKED ISSUE #557: Conversion from 64-bit Number to BigInt is lossy
**Describe the bug**
While solving issue https://github.com/tweag/asterius/issues/373 (PR https://github.com/tweag/asterius/pull/542), we found out that the conversion of 64-bit JavaScript `Number`s into `BigInt`s is lossy. Though somewhat expected [1], this has repercussions for Asterius, since the 64-bit versions of `memset` do not work as expected when contents above `MAX_SAFE_INTEGER` are used (this affects types that are represented using 64 bits such as `Word`, `Word64`, and `Ptr`). For example, running this piece of code
```haskell
do
  dst <- newPinnedByteArray 32
  setByteArray dst (3*8) 1 (0xcccccccccccccccc :: Word64)
  w <- readByteArray dst 3
  print (w == (0xcccccccccccccccc :: Word64))
```
unintuitively prints `False`.

[1]: The maximum integer that can be safely represented in JavaScript is `Number.MAX_SAFE_INTEGER = 2^53 - 1`, because JavaScript uses IEEE754 double precision floating point number representation (see [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER)).

**To Reproduce**
The easiest way to reproduce the issue is to run:
```
stack test asterius:primitive
```
This test uses all Haskell variants of `memset` on a given array, and checks that storing (using `memset`) and retrieving should yield the same result. Hence, it is expected that running the test prints only `True`s (each `False` is a case where this above property is violated).

**Expected behavior**
Running `stack test asterius:primitive` should only print `ByteArray`s and `True`.

**Environment**
 - OS name + version: Ubuntu 18.04.4 LTS (bionic)
 - GHC version: 8.8.3
 - Cabal version: 3.0.1.0
 - Version of Asterius: current HEAD (https://github.com/tweag/asterius/commit/5ad234e1f486cbbb618abb42d73811d8b10e255c)

**Additional context**

The conversion from `Number` to `BigInt` happens in `asterius/rts/rts.memory.mjs` where a JS version of `memset` is implemented:

```javascript
memset(_dst, c, n, size = 1) {
  // We only allow 1, 2, 4, 8. Any other size should get a runtime error.
  const ty = {
    1 : Uint8Array,
    2 : Uint16Array,
    4 : Uint32Array,
    8 : BigUint64Array
  }; 
  const buf = this.expose(_dst, n, ty[size]);

  if (size === 8) {
    // TODO: The conversion BigInt(c) is lossy. Numbers are represented as 
    // IEEE754 double precision floating point numbers, for which the maximum
    // (representable) safe integer in JavaScript is (Number.MAX_SAFE_INTEGER
    // = 2^53 - 1).
    buf.fill(BigInt(c));
  } else {
    buf.fill(c);
  }  
}
```
The input values are always `Number`s, even when we want them to represent 64-bit integers, so the conversion `BigInt(c)` is unavoidable.

### LINKED ISSUE #718: Use BigInt for i64 FFI and remove existing i64/f64 casting tricks
WebAssembly MVP prohibits `i64` as an FFI argument/result type, and we had to work around the restriction. Previously, the major way is performing `i64`/`f64` casting at the wasm/js boundary, which bears the risk of undetected precision loss. There has been a WebAssembly [proposal](https://github.com/WebAssembly/JS-BigInt-integration) for enabling wasm `i64` as an FFI type using JavaScript `BigInt`, but we held off using that since it wasn't enabled in major browsers.

This is now no longer the case: 

* The proposal proceeded into WebAssembly main spec
* V8 [announced](https://v8.dev/blog/v8-release-85#support-for-js-bigints) in a recent release that the feature has been shipped
* According to [Wasm Feature Detect](https://wasm-feature-detect.surma.technology/), Firefox nightly and Chrome dev supported this feature

It is safe to assume that in the near future, this feature will be available on major platforms, and we can take advantage of it to remove existing `i64`/`f64` casting on the FFI boundary.

* [ ] Enable the `--experimental-wasm-bigint` flag for `node` since this isn't on by default on `node `yet
* [ ] Refactor the JavaScript runtime and the builtin functions, use `i64` for imports/exports and remove all existing casting logic
* [ ] Refactor the JSFFI code generator, so that the JSFFI imports/exports use `i64` in the wasm import/export type signature, and do not perform casting in the wrapper function
* [ ] Mention in documentation about the new WebAssembly feature we use

**Describe alternatives you've considered**
The main alternative approach is preserving a linear memory region as the "value stack" for passing arguments/results across FFI boundary. It's the mechanism used by `wasm-bindgen` in rust. We may implement this in the future, but it takes a lot more work, thus we should emphasize on getting rid of i64/f64 casting for now to increase runtime robustness.

gkaracha:
Work in progress. When finished, this should resolve #718 (and also fix #557).


gkaracha:
@TerrorJack surprisingly (to me) running test `primitive` with this PR currently hangs:
```
$ stack test asterius:primitive
asterius> test (suite: primitive)

[INFO] Compiling test/primitive/primitive.hs to WebAssembly
[INFO] Converting linked IR to binaryen IR
[INFO] Running binaryen optimization
[INFO] Validating binaryen IR
[INFO] Writing WebAssembly binary to "test/primitive/primitive.wasm"
[INFO] Writing JavaScript runtime modules to "test/primitive"
[INFO] Writing JavaScript loader module to "test/primitive/primitive.wasm.mjs"
[INFO] Writing JavaScript req module to "test/primitive/primitive.req.mjs"
[INFO] Writing JavaScript entry module to "test/primitive/primitive.mjs"
[INFO] Running test/primitive/primitive.mjs

```
Could you test whether you can reproduce this behavior locally too?

Issue #697, opened 2020-06-22
GitHub check ghc-testsuite is fragile


gkaracha:
**Describe the bug**
On both `push` and `pull_request`, the GitHub check `ghc-testsuite` stochastically (and frequently) fails, without a clear indication of the source of the failure.
 
**To Reproduce**
Commit and push any change.

**Expected behavior**
The check should either (a) pass, or (b) fail and provide the log output as an artifact.

**Environment**
 - OS name + version: not applicable
 - Version of the code: current HEAD (d946f2d41e68983174ef32d3b58d739b373f7591)

PR #724, opened 2020-08-03
Fix ghc-testsuite major regression

TerrorJack:
`ghc-testsuite` has regressed (~600 failures); the regressed tests would produce an empty `stdout`/`stderr`, and the root cause is the `main` `IO` action in compiled modules haven't been fully executed and returned before we attempt to read `stdout`/`stderr`.

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -161,7 +161,9 @@ jobs:
 
       - name: ghc-testsuite
         run: |
-          GHCRTS=-N2 stack test asterius:ghc-testsuite --test-arguments="-j2 --timeout=300s" || true
+          export GHCRTS=-N2
+          export ASTERIUS_GHC_TESTSUITE_TIMEOUT=300
+          stack test asterius:ghc-testsuite --test-arguments="-j2" || true
 
       - name: upload-artifact
         uses: actions/upload-artifact@v2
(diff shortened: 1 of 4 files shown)


TerrorJack:
I figured out the cause of #697 btw: we set `--timeout=300s`, and the timeout mechanism of `tasty` (based on `unbounded-delays`) is not working for certain cases (e.g. `T5536`). This needs to be fixed as well.

gkaracha:
> I figured out the cause of #697 btw: we set `--timeout=300s`, and the timeout mechanism of `tasty` (based on `unbounded-delays`) is not working for certain cases (e.g. `T5536`). This needs to be fixed as well.

Nice! I wonder whether it's a good idea to increase the timeout for all tests, or if it would be better to put the performance tests `perf` into a separate job, which has a more lax timeout?

TerrorJack:
@gkaracha Bumping the timeout number won't help, since for some reason the timeout mechanism doesn't work at all for certain cases. This sounds a bit frightening but I'm not really surprised, since timeouts require interrupting `IO` actions via async exceptions, and async exceptions are fragile due to misused masking, unsafe FFI calls, etc..

I have implemented our own version of `timeout` and use an environment variable to control the `ghc-testsuite` for now. The custom `timeout` is a lot less efficient than `base`/`unbounded-delays` implementation, but it guarantees that the returned `IO` action immediately throws once the time limit exceeded, even if the wrapped `IO` action is not really terminated. This seems to work locally for the stubborn `perf` cases, let's see how it goes on CI.

TerrorJack:
The custom `timeout` now works for all tests; there are other recent problems exposed in `ghc-testsuite` which shall be fixed in other PRs.

Issue #722, opened 2020-08-02
"RuntimeError: unreachable"


Lemmih:
**Describe the bug**
Running this code fails with 'RuntimeError: unreachable':
```
{-# LANGUAGE PackageImports #-}
module Main where

import Graphics.SvgTree
import Graphics.SvgTree.Printer
import "lens" Control.Lens

main :: IO ()
main = putStrLn $ ppTree (CircleTree $ Circle mempty (Num 0, Num 0) (Num 2))
```
```
root@ac07adc0d6f9:/workspace# ahc-link --input-hs test.hs --run
[INFO] Compiling test.hs to WebAssembly
[INFO] Converting linked IR to binaryen IR
[INFO] Running binaryen optimization
[INFO] Validating binaryen IR
[INFO] Writing WebAssembly binary to "./test.wasm"
[INFO] Writing JavaScript runtime modules to "."
[INFO] Writing JavaScript loader module to "./test.wasm.mjs"
[INFO] Writing JavaScript req module to "./test.req.mjs"
[INFO] Writing JavaScript entry module to "./test.mjs"
[INFO] Running ./test.mjs
test: JSException "RuntimeError: unreachable\n    at <anonymous>:wasm-function[5314]:0x1320cb\n    at <anonymous>:wasm-function[12289]:0x2db0f9\n    at scheduleTSO (<anonymous>:wasm-function[12290]:0x2db126)\n    at Scheduler.tick (file:///workspace/rts.scheduler.mjs:346:22)\n    at Immediate.<anonymous> (file:///workspace/rts.scheduler.mjs:381:29)\n    at processImmediate (internal/timers.js:458:21)"
```
Running the same code but with a circle radius of '0' works fine:
```
main = putStrLn $ ppTree (CircleTree $ Circle mempty (Num 0, Num 0) (Num 2))
```
```
root@ac07adc0d6f9:/workspace# ahc-link --input-hs test.hs --run
[INFO] Compiling test.hs to WebAssembly
[INFO] Converting linked IR to binaryen IR
[INFO] Running binaryen optimization
[INFO] Validating binaryen IR
[INFO] Writing WebAssembly binary to "./test.wasm"
[INFO] Writing JavaScript runtime modules to "."
[INFO] Writing JavaScript loader module to "./test.wasm.mjs"
[INFO] Writing JavaScript req module to "./test.req.mjs"
[INFO] Writing JavaScript entry module to "./test.mjs"
[INFO] Running ./test.mjs
<circle />
[...]

PR #725, opened 2020-08-04
Make -DASTERIUS a wired-in CPP flag

TerrorJack:
This PR makes `-DASTERIUS` a wired-in CPP flag for `ahc`. Asterius users shall be able to use the `ASTERIUS` CPP macro to determine whether the current module is being compiled by asterius, without specifying additional flags to `ahc-cabal`, `ahc-link`, etc. Relevant: #722

### DIFF
--- a/asterius/boot.sh
+++ b/asterius/boot.sh
@@ -17,7 +17,7 @@ cd ..
 
 cd base
 autoreconf -i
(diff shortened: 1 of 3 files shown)


Lemmih:
Woo!

PR #732, opened 2020-08-11
Add a section about code formatting in hacking.md

gkaracha:


gkaracha:
@TerrorJack do we use any other code formatters other than the two I mention? I think it's good to have all this information in one place.

TerrorJack:
We only use these two at the moment. Well, I tend to run `brittany` before `ormolu` to get a more canonical result, but I don't intend to enforce this on other contributors as well :stuck_out_tongue:

gkaracha:
> We only use these two at the moment. Well, I tend to run `brittany` before `ormolu` to get a more canonical result, but I don't intend to enforce this on other contributors as well :stuck_out_tongue: 

:joy: Good, even if we could put it as "advice" I definitely won't follow it, way too much work :stuck_out_tongue:

PR #736, opened 2020-08-11
Add support for Position Independent Code (PIC) - Part A

### LINKED ISSUE #723: PIC(Position Independent Code) support
When #658 is implemented, dlmalloc in `wasi-libc` will take over control of the linear memory. In order to layout static data segments (e.g. info tables, static closures) in the memory, we need to call `malloc` to allocate the static data region. The returned address is only known at runtime.

This indicates that we need to generate PIC(Position Independent Code). Previously, at link-time, each symbol gets rewritten to its absolute address. Now, each symbol needs to be rewritten to a base address plus an offset. The offset is still known at link-time, but the base address is only known at run-time. The result of calling `malloc` will become an imported immutable global value, to be used as the base address. The combination of imported immutable global value plus a constant offset is permitted in function code and memory/table initialization.

One remaining problem is: what if the content of a data segment refers to a symbol? Since data segment contents must be vanilla buffers instead of wasm constant expressions, we can't directly put the base address + offset expression there. The solution is: the linker generates a function which corrects the data segment contents. That function is called only once when initializing the runtime.

* Introduce the `env.__memory_base` `i32` global import
* All data addresses are rewritten to `TAG(env.__memory_base + offset)`. `offset` is also `i32`, and `TAG()` converts an `i32` physical address to our `i64` virtual address.
* At link-time, track all symbols in data segments, generate a `__wasm_apply_relocs` function which corrects the symbol contents given `env.__memory_base`
* Fix the symbol table logic in the runtime. All symbol table entries are offsets in the generated JavaScript artifact, and they must be rewritten to `TAG(env.__memory_base + offset)` as well.
* Apply the same to functions, using `env.__table_base` as the table offset
* Set the offsets to constant 1K in the runtime for now, before we actually call `malloc` to handle the memory.

**Describe alternatives you've considered**
Use `lld`'s `--global-base` flag to reserve a large unused linear memory region for static data when producing the `wasi-libc` module. The static data segments go to that region.

The main problem is: a different `wasi-libc` module must be linked each time we link an asterius module. Complicates the build system significantly. Given PIC will be required for dynamic loading anyway, might as well do it the proper way.

**Additional context**
https://github.com/WebAssembly/tool-conventions/blob/master/DynamicLinking.md

gkaracha:
This PR is the first part of addressing #723.

* Add two `WebAssembly.Global`s, `__asterius_memory_base` (statics) and `__asterius_table_base` (functions).
* Disentangle `req.exportsStatic` into `req.functionsExportsStatic` and `req.staticsExportsStatic`; since the base address is different for each namespace, we need to be able to tell them apart.
* Disentangle `req.symbolTable` into `req.functionsOffsetTable` and `req.staticsOffsetTable` for the same reason.
* Define a `SymbolTable` object, to opaquely lookup addresses of statics and functions, via `addressOf`. Now we get the offsets from the `.req.mjs` and methods `getMemoryBase()` and `getTableBase()` give the base addresses.
* Change the elaboration of (function) `Symbol`s so that it utilizes the (currently zero) base address read from the global, for both the `binaryen` and the `wasm-toolkit` backends.


TerrorJack:
Would be nice to factor out "disentangling the data/function symbol map in the backends" as a separate PR, since it is not strictly related to PIC and can be merged on its own. Other parts lgtm for now.

gkaracha:
> Would be nice to factor out "disentangling the data/function symbol map in the backends" as a separate PR, since it is not strictly related to PIC and can be merged on its own.

Good idea, I'll do that :+1: 

> Other parts lgtm for now.

About the next step: currently `genReq` generates the contents of the `.req.mjs` file, and it dumps the complete and absolute Asterius addresses (tag + actual) into the file at the moment (in `symbolTable`). How would you suggest we go about changing these contents dynamically?

TerrorJack:
> About the next step: currently genReq generates the contents of the .req.mjs file, and it dumps the complete and absolute Asterius addresses (tag + actual) into the file at the moment (in symbolTable). How would you suggest we go about changing these contents dynamically?

* The interface of querying a symbol via `symbolTable.someSymbol` must be changed to something like `symbolTable.get("someSymbol")`. `symbolTable` used to just be a simple dict, now it needs to be yet another runtime component exposing a method to calculate the address.
* The `symbolTable` can be initialized via a mapping from symbol strings to offsets (rather than absolute addresses before). The `get` method converts from the offset to the absolute address.

TerrorJack:
We haven't began working on the `__wasm_apply_relocs` generation logic yet, but there's one thing worth noting here: `__wasm_apply_relocs` function size will grow as the final `.wasm` size grows, and it's likely to overflow the maximum wasm function size when the linker output is big enough (e.g. `pandoc`).

The solution would be: `__wasm_apply_relocs` itself would call several splitted functions, each function contains a part of the store instructions used to correct the data segments. We shouldn't make a reloc function for each data segment containing symbol, since that'll bloat the final `.wasm` size; but splitting will still be necessary. The linker shall hard-code a threshold value indicating the maximum number of store instructions allowed in such a sub function.

Ref:

https://reviews.llvm.org/D75703
https://source.chromium.org/chromium/chromium/src/+/master:v8/src/wasm/wasm-limits.h;l=39;bpv=1;bpt=1?q=limit

TerrorJack:
Also, once the `addressOf` related stuff work, it would be nice to factor this part out as a standalone PR too, given the growing size of the current PR.

asterius/src/Asterius/Backends/Binaryen.hs
```diff
@@ -514,7 +515,16 @@ marshalExpression e = case e of
...
         | Just x <- SM.lookup unresolvedSymbol func_sym_map ->
-          lift $ Binaryen.constInt64 m $ x + fromIntegral symbolOffset
+          let base =
+                GetGlobal
+                  { globalSymbol = "__asterius_table_base",
+                    valueType = I32
+                  }
+           in marshalExpression $
+                addInt64
+                  (extendUInt32 base) -- TODO: or extendSInt32? Not sure which has the correct semantics.
```
TerrorJack, line 525: * `extendUInt32` is correct here, since we never handle negative base values
* We can perform `extendUInt32` on the result of adding the offset onto base; the offset would need to be an `i32` value then.

asterius/src/Asterius/Backends/Binaryen.hs
```diff
@@ -514,7 +515,16 @@ marshalExpression e = case e of
...
+                  (constI64 $ fromIntegral x + symbolOffset)
+                  -- TODO: fromIntegral performs potentially lossy downcast from In64 to Int here
```
TerrorJack, line 527: This is safe here, since we only work with 64-bit asterius.

asterius/src/Asterius/Backends/Binaryen.hs
gkaracha: > * `extendUInt32` is correct here, since we never handle negative base values

Right :+1: 

> * We can perform `extendUInt32` on the result of adding the offset onto base; the offset would need to be an `i32` value then.

Hmmm, not sure what you mean by this. How would you write this expression instead?

TerrorJack: ``extendUInt32 (base `addInt32` fromIntegral (x + symbolOffset))``

This would also mean that we need to represent the symbol map with `SymbolMap Int` rather than the previous `SymbolMap Int32`. But changing the symbol map type can go in later PRs I think

gkaracha: I see. Well, currently the `x` is an `Int64`, using the upper 32 bits for the tag so it really is more than `Int32`.

TerrorJack (commented):
Is it too much work to remove the currently unused `__wasm_apply_relocs` logic? If so we can leave it as it is for now, but still better to leave out irrelevant logic in a PR.

gkaracha:
> Is it too much work to remove the currently unused `__wasm_apply_relocs` logic? If so we can leave it as it is for now, but still better to leave out irrelevant logic in a PR.

Yea, that shouldn't be an issue. Lemme revert that real quick.

TerrorJack:
Let's just revert https://github.com/tweag/asterius/commit/02054100a0dedc524b1ec74a0002af27442daf5a and merge once CI is green again.

The cause of the crash is: some code reads a Cmm function address from the data segments and jump to it. The problem is, since we haven't implemented the `__wasm_apply_relocs` logic yet, the data segments don't contain the correct function address, but the offset value instead. Before that commit, the base value is `0` and the offset happens to be the correct address.

gkaracha:
> Let's just revert [0205410](https://github.com/tweag/asterius/commit/02054100a0dedc524b1ec74a0002af27442daf5a) and merge once CI is green again.

If you think that this PR does not belong on `master` as is (I mean if it only makes sense when Part B has landed too), I could work on Part B in a separate PR built on top of this one. Then we can merge them both when Part B is ready. Would that perhaps be better?

TerrorJack:
Let's merge this one before Part B is available. This PR involves changes in both runtime & compiler, and gets easily bit-rotten when working on other parts.

gkaracha:
> This PR involves changes in both runtime & compiler, and gets easily bit-rotten when working on other parts.

True. OK then, if the CI passes feel free to merge it; I'll get going with Part B :slightly_smiling_face:

gkaracha:
@TerrorJack no idea why `test-cabal` fails, but the rest seem to pass fine; I think that this failure does not relate to the PR.

PR #750, opened 2020-08-24
Add support for Position Independent Code (PIC) - Part B

gkaracha:
This PR is the second part of addressing https://github.com/tweag/asterius/issues/723.

* Introduce the `--pic` flag for `ahc-ld` and `ahc-link`. When `--pic` is on, we generate position independent code (almost, see below for remaining tasks). The flag is also available in the rts, as well as the default values for `tableBase` and `memoryBase`.
* Re-add the `__asterius_table_base` and `__asterius_memory_base` globals in the rts. Their value can now depend on whether the `--pic` flag is on.
* Create `__asterius_spt_entries` in the runtime from the corresponding offsets, instead of doing so at compile-time.
* Create `__asterius_info_tables` in the runtime from the corresponding offsets, instead of doing so at compile-time.
* Adapt the creation of data segments to create an accompanying relocation function `__wasm_apply_relocs`, which is called in the runtime when `--pic` is on. If `--pic` is off, `__wasm_apply_relocs` is still in scope, but it's implementation is empty (no-op).
* Create the symbol table in the rts using the two (now offset) tables **and** the base addresses, known at runtime.
* Introduce two more CI jobs for testing `--pic` (one for `binaryen` and one for `wasm-toolkit`).
* In order to circumvent the restriction that the `offset` in `DataSegment` can only be a [constant expression](https://webassembly.github.io/spec/core/valid/instructions.html#valid-constant), if `--pic` is on we squash all data segments into one; in this case the offset is simply a `GetGlobal` (for the `__asterius_memory_base`).
* In order to avoid overflowing the maximum wasm function size (when the linker output is big enough), as described in https://reviews.llvm.org/D75703#1926884, we store all offsets that `__wasm_apply_relocs` needs to change in two data segments.

Remaining TODOs:
* Add an entry in [the docs](https://github.com/tweag/asterius/blob/master/docs/ahc-link.md) about `--pic`, once PIC is really operational.


gkaracha:
@TerrorJack can you have a look at the current state of the PR? It looks fine to me but when I change the base address to something that is non-zero I get [the same error as before](https://github.com/tweag/asterius/pull/736#issuecomment-678357510). I must be missing something.

asterius/rts/rts.mjs
```diff
@@ -220,6 +220,7 @@ export async function newAsteriusInstance(req) {
 
     Object.assign(__asterius_exports, __asterius_wasm_instance.exports);
     __asterius_exports.hs_init();
+    __asterius_exports.__wasm_apply_relocs();
```
TerrorJack, line 223: We call `__wasm_apply_relocs` as an exported function, but it seems it hasn't been declared as a function export anywhere?

TerrorJack (changes requested):

asterius/rts/rts.mjs
gkaracha: It is declared as part of `rtsFunctionExports` here: https://github.com/tweag/asterius/pull/750/files#diff-52a18eb8b68ef165a247a8fc8ebf536bR683

TerrorJack (changes requested):
There may be other details I've yet to check, here're some points to address for now:

* `makeDataSymbolTable`/`makeFunctionSymbolTable` used to take a "starting address" parameter, and by setting the starting address to an already tagged address we ensure all the produced addresses are already tagged. For proper PIC however, we should take away this parameter and produce the results starting from `0`. And we should rename "symbol table" to "offset table" to avoid ambiguity.
* Since both the offsets and the imported memory/table base values are `i32`s, the correct symbol value should be `TAG(base+offset)`, and `TAG()` should first convert `i32` to `i64` before adding the tag.
* `__wasm_apply_relocs` must be called as soon as the `WebAssembly.Instance` object is created, even before `hs_init`. I'll also need to check that the initialization of various rts components don't actually read into the yet invalid data segments yet.

gkaracha:
Nice points, I'll get to it right away :+1:

gkaracha:
@TerrorJack since all the tables now contain 32-bit offsets, I'd like to switch from `SM.SymbolMap Int64` to `SM.SymbolMap Int32`, but I'm a little worried that it'll bloat the diff. What do you think?

TerrorJack:
@gkaracha Let's go for `Int32`. The diff is already large anyway.

gkaracha:
> The diff is already large anyway.

True that; I'll switch to `Int32` then :+1:

TerrorJack:
@gkaracha Would you do a merge for this branch? After that I can take another look today.

gkaracha:
> @gkaracha Would you do a merge for this branch? After that I can take another look today.

Definitely. It will take a little to rebase this beast, but I hope that after rebasing we'll find more opportunities for pieces to factor out. I'll ping once it's done.

gkaracha:
@TerrorJack you can have another look now if you want. Locally since I corrected the implementation of `makeWasmApplyRelocs` I have not been able to compile anything; binaryen optimization hangs forever. Actually, it already took a long time to complete earlier: small changes in `makeWasmApplyRelocs` slow down binaryen optimization a lot for some reason. I thought that the size of the generated `__wasm_apply_relocs` function was massive, but that's not the case (I tried out bloaty and it wasn't that big).

TerrorJack:
@gkaracha Thanks, will take a look right away. And don't forget there's `--optimize-level=0 --shrink-level=0` you can pass to `ahc-link` to temporarily disable all `binaryen` optimization.

gkaracha:
> @gkaracha Thanks, will take a look right away.

:+1: 

> And don't forget there's `--optimize-level=0 --shrink-level=0` you can pass to `ahc-link` to temporarily disable all `binaryen` optimization.

So true, I didn't even think of that :facepalm: I'll check it out, see if we can at least get a result.

gkaracha:
Well not much has changed, really, we still get the thread stopping reason error (FWIW, the number is different than it was before revising `makeWasmApplyRelocs`):
```
RuntimeError: returnFFIPromise: unsupported thread stopping reason 141472
```

asterius/src/Asterius/Passes/DataOffsetTable.hs
```diff
@@ -0,0 +1,152 @@
...
+  SM.SymbolMap Int64 ->
+  (Int32, Set.Set Int32, Set.Set Int32) ->
+  AsteriusStatic ->
+  ((Int32, Set.Set Int32, Set.Set Int32), Bag DataSegment)
+makeSegment fn_off_map ss_off_map (current_off, fn_meta, ss_meta) static = case static of
+  SymbolStatic sym o
+    | Just off <- SM.lookup sym fn_off_map ->
+      let off_to_store = off + fromIntegral o
+       in ( (next_off, current_off `Set.insert` fn_meta, ss_meta),
+            unitBag DataSegment {content = encodeStorable off_to_store, offset = current_off}
```
TerrorJack, line 105: Looks like another critical error to me. In the `DataSegment`s we produce in this module, the `offset`s start from `0` right? (Based on `makeDataOffsetTable`). Well, they really should be `memory_base + offset`. And this also implies that the `DataSegment` should make the `offset` field an `Expression` type.

TerrorJack (changes requested):

asterius/src/Asterius/Passes/DataOffsetTable.hs
gkaracha: Ah, indeed. I'll do the generalization in a separate PR, like for the `FunctionTable`.

TerrorJack: #777 is merged, would you solve the conflict here and see if the corrected offsets solve the crashes?

gkaracha: Will do.

gkaracha:
@TerrorJack Well, interestingly, now it gives a binaryen validation error:
```
[wasm-validator error in module] unexpected false: segment offset should be reasonable, on 
[i32] (i32.add
 [i32] (global.get $__asterius_memory_base)
 [i32] (i32.const 148136)
)
```
Have not seen such an error before :thinking: We can continue the investigation next week.

TerrorJack:
@gkaracha I was wrong about what "constant expression" meant in wasm spec. I thought it should be any expression which can be evaluated to a constant, but according to https://webassembly.github.io/spec/core/valid/instructions.html#valid-constant, it should either be a `global.get` or a `const` and can't be composed in any other way. And `binaryen` implemented this check in validation. I need to rethink how to handle data segments in the PIC context here.

Another bad news is: even if I hard-code the `offset`s to `ConstI32 (1024+current_off)`, it would still crash. So there are other places to look at. But have a nice weekend anyway!

gkaracha:
> @gkaracha I was wrong about what "constant expression" meant in wasm spec. I thought it should be any expression which can be evaluated to a constant, but according to https://webassembly.github.io/spec/core/valid/instructions.html#valid-constant, it should either be a `global.get` or a `const` and can't be composed in any other way. And `binaryen` implemented this check in validation. I need to rethink how to handle data segments in the PIC context here.
> 
> Another bad news is: even if I hard-code the `offset`s to `ConstI32 (1024+current_off)`, it would still crash. So there are other places to look at. But have a nice weekend anyway!

Ah, I see! Well, for now I will set it to `ConstI32 (1024+current_off)` regardless, so that we can proceed with debugging, and indeed we'll have to find a solution for this one eventually.

gkaracha:
@TerrorJack I'll remove the duplication in `.github/workflows/pipeline.yml` once https://github.com/commercialhaskell/stack/issues/5387 is resolved. I also want to experiment locally with different values for `defaultTableBase` and `defaultMemoryBase` before undrafting the PR. Do you have any other suggestions/criticisms I could work on in the meantime?

gkaracha:
Also, I tried running all of `test-binaryen` with the following configurations
```
defaultTableBase =    1, defaultMemoryBase = 1024
defaultTableBase = 3517, defaultMemoryBase = 1024
defaultTableBase = 3517, defaultMemoryBase = 4096
```
and they all work fine! :grin:

gkaracha:
Now that I look at it, maybe it's best to only call `i.exports.__wasm_apply_relocs()` if `--pic` is on, even if otherwise it's implementation is empty. The less we have to pay if `--pic` is off the better.

gkaracha:
@TerrorJack yes, I think this can work :+1: For what it's worth though, this means that withing `resolveAsteriusModule` we now have to patch `ss_off_map` and `fn_off_map`, on top of the patched `final_m`. At least all this complication will remain within `resolveAsteriusModule`.

On a side note, I looked into the two remaining `TODO`s—`tableSlots` and `memoryMBlocks` computed at compile time—and I see no way we compute either at runtime. They are used by `BinaryenSetFunctionTable` and `BinaryenSetMemory`, respectively, and these both expect an `Index`, not an `Expression` :confused:

TerrorJack:
> At least all this complication will remain within resolveAsteriusModule
More specifically I think the extra complication is in `DataOffsetTable.hs`

> They are used by BinaryenSetFunctionTable and BinaryenSetMemory, respectively, and these both expect an Index, not an Expression

I don't get it; `BinaryenSetFunctionTable` takes a `BinaryenExpressionRef` offset parameter, and we already changed our `FunctionTable` `offset` to `Expression` right? Same for memory

gkaracha:
> More specifically I think the extra complication is in `DataOffsetTable.hs`

The extra computation yes, but `resolveAsteriusModule` will have to patch the offset maps.

> 
> > They are used by BinaryenSetFunctionTable and BinaryenSetMemory, respectively, and these both expect an Index, not an Expression
> 
> I don't get it; `BinaryenSetFunctionTable` takes a `BinaryenExpressionRef` offset parameter, and we already changed our `FunctionTable` `offset` to `Expression` right? Same for memory

Sorry, I wasn't clear enough. The `FunctionTable` and `DataSegment` are generalized indeed so that their offsets are `Expression`s. What I was talking about was the `table_slots` (computed in `resolveAsteriusModule` and stored into the resulting `Module`), which are then used by `marshalFunctionTable` (in `Asterius.Backends.Binaryen`), which calls
```haskell
foreign import ccall unsafe "BinaryenSetFunctionTable"
  setFunctionTable ::
    Module ->
    Index ->  -- this is the tableSlots argument
    Index ->
    Ptr (Ptr CChar) ->
    Index ->
    Expression ->
    IO ()
```

gkaracha:
(see the two `-- TODO: make dynamic` in `resolveAsteriusModule` and follow their trail to the backend for the above)

TerrorJack:
@gkaracha Here's a detailed writeup on the semantics of the index values and how to deal with them:

* `WebAssembly.Memory` and `WebAssembly.Table` are created in JavaScript and imported into WebAssembly. When creating them, we need to specify the initial sizes (number of pages or table slots)
* When we specify memory/table imports in WebAssembly, we also need to specify the minimal sizes (which are the `Index` parameters we're talking about).
* According to https://webassembly.github.io/spec/core/exec/modules.html#import-matching, at runtime, the engine will check if the imported memory/table size is at least the size specified in the wasm imports, and throw an error if not.
* At link-time, by the time we have the finalized symbol offset maps, we know at least how much memory (in pages) and table slots will be required. Let's call these sizes `s0`. `s0` does not include the 1K memory padding or the 1 reserved table slot.
* The minimal sizes we pass to `BinaryenSetMemory` or `BinaryenSetFunctionTable` should be `s0`.
* In JavaScript, when creating `WebAssembly.Memory` and `WebAssembly.Table`, the initial sizes should be `s0` plus the 1K/1 reserved space.

gkaracha:
@TerrorJack Thanks for the explanation, I understand what I have to do :ok_hand: Well, this is great news; if my calculations are correct, after fixing these two we should **really** have position independent code :crossed_fingers:

gkaracha:
So, 98eacf2 fixes a serious omission: passing `pic` to `ahc` in `ahcLink`.

Now that all the logic is there, only thing remaining to do is fix the issues in the relocation function (in `Asterius.Passes.DataOffsetTable`) that give us out-of-bounds errors:
```
(node:22321) UnhandledPromiseRejectionWarning: RuntimeError: memory access out of bounds
    at __wasm_apply_relocs (<anonymous>:wasm-function[55]:0x3104)
    at file:///home/skull/tweag/a/test/rts.mjs:239:17
```

TerrorJack:
@gkaracha There is an earlier revision that doesn't generate reloc segments and instead encodes reloc info as explicit load/store instructions in `__wasm_apply_relocs` right? Might be worth it to try committing the `--optl--pic` fix there to see if things work.

gkaracha:
Good idea @TerrorJack, at least that implementation was simpler and easier to debug.

TerrorJack:
@gkaracha What's the status quo? Has all cleanup been done yet? Also would be nice to do a round of grep for todos before we hit the merge button.

gkaracha:
> @gkaracha What's the status quo? Has all cleanup been done yet? Also would be nice to do a round of grep for todos before we hit the merge button.

Going through all the changes now, adding comments where needed, and fixing small TODOs. I'll ping you when I'm done (later today).

gkaracha:
@TerrorJack this should be good to go; let's just wait for CI to be green

TerrorJack (approved):

PR #763, opened 2020-08-26
Wasm implementations of hs_bswap and ntohs functions

### LINKED ISSUE #756: Missing `ntohs` functions
**Describe the bug**
The runtime doesn't implement [`ntohs`](https://linux.die.net/man/3/ntohs) functions. They're used by [`Network.Socket.Types`](https://github.com/haskell/network/blob/master/Network/Socket/Types.hsc) in the `network` package, specifically, the `Show` instance of the `SockAddr` type uses them. Attempting to use it will cause a runtime crash.

EDIT: The [`Show SockAddr` instance](https://hackage.haskell.org/package/network-3.1.1.1/docs/src/Network.Socket.Info.html#line-438) also depends on `c_getnameinfo`; this seems to be the source of the error even with `ntohs` implementations available. This issue shall thus be addressed when we integrate a C toolchain.

**To Reproduce**
Use the `Show` instance of `SockAddr` at runtime. Or transitively, the `Show` instance of `Request` from the `wai` package.

**Expected behavior**
It shouldn't crash.

**Environment**
 - OS name + version: Docker dev image
 - Version of the code: latest `master` revision

**Additional context**

gkaracha:
Closes #756. This PR

Adds `Asterius.Builtins.Endianness` with implementations for
* `hs_bswap16`, `hs_bswap32`, and `hs_bswap64`.
* `htonl`, `htons`, `ntohl`, and `ntohs`, based on the `hs_bswap*` functions. Wasm uses little-endian while network order is big-endian; converting from one to the other simply means byte-swapping.

Also
* Adds a small unit test (`endianness`).
* Implements translations for [`CallishMachOp`](https://haskell-code-explorer.mfix.io/package/ghc-8.6.5/show/cmm/CmmMachOp.hs#L523) `MO_BSwap`, using `hs_bswap16`, `hs_bswap32`, and `hs_bswap64`.
* Adds a few comments about remaining unimplemented translations in `Asterius.CodeGen`.

### DIFF
--- a/asterius/test/endianness/endianness.hs
+++ b/asterius/test/endianness/endianness.hs
@@ -1,0 +1,6 @@
+
+-- import Network.Socket
+
+main :: IO ()
+main = putStrLn "This test is under construction!"
+
--- a/asterius/test/endianness/endianness.hs
+++ b/asterius/test/endianness/endianness.hs
@@ -4,21 +4,36 @@ import Data.Word
 
 main :: IO ()
 main = do
-  do let x = 10
-     print x
-     print (ntohs x)
-  do let x = 10
-     print x
-     print (ntohl x)
-  do let x = 10
-     print x
-     print (htons x)
-  do let x = 10
-     print x
-     print (htonl x)
+  do let x = 0xABCD
+     let y = 0xCDAB
+     putStr "ntohs: " >> print (ntohs x == y && x == ntohs y)
+  do let x = 0x56789ABC
+     let y = 0xBC9A7856
+     putStr "ntohl: " >> print (ntohl x == y && x == ntohl y)
+  do let x = 0xABCD
+     let y = 0xCDAB
+     putStr "htons: " >> print (htons x == y && x == htons y)
+  do let x = 0x56789ABC
+     let y = 0xBC9A7856
+     putStr "htonl: " >> print (htonl x == y && x == htonl y)
+  --
+  do let x = 0xABCD
+     let y = 0xCDAB
+     putStr "hs_bswap16: " >> print (hs_bswap16 x == y && x == hs_bswap16 y)
+  do let x = 0x56789ABC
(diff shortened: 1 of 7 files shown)


TerrorJack:
Might as well implement the `hs_bswap` functions [here](https://github.com/tweag/asterius/blob/master/ghc-toolkit/boot-libs/ghc-prim/cbits/bswap.c) first, use those for `ntohs`. Once `hs_bswap` works we can fix code generation of the byteswap primops as well (handle `MO_BSwap` in `Asterius.CodeGen`)

asterius/test/endianness/endianness.hs
```diff
@@ -0,0 +1,39 @@
+
+-- import Network.Socket -- Well, it's not booted, this one, so we cannot test it :(
```
TerrorJack, line 2: See https://github.com/tweag/asterius/blob/master/.github/workflows/pipeline.yml#L86, `network` is actually booted in CI.

TerrorJack (changes requested):

asterius/test/endianness/endianness.hs
gkaracha: It is? Locally I could not find it. Maybe `./utils/reboot.sh` is out-of-sync.

gkaracha: > See https://github.com/tweag/asterius/blob/master/.github/workflows/pipeline.yml#L86, `network` is actually booted in CI.

Indeed, running

> . ./.envrc
> ahc-cabal v1-update 
> pushd ghc-toolkit/boot-libs
> ahc-cabal v1-install -j2 wai-middleware-caching
> popd

locally worked just fine. I'll add a test for `Network.Socket` as well, to make sure #756 is addressed :+1:

TerrorJack (approved):

Issue #332, opened 2019-10-20
Obtaining unit ids in a principled way


TerrorJack:
## Background

Asterius needs to know the symbols of certain entities (closures & info tables), for the use of rts js code and generated builtin functions. In the past, those entities were all from wired-in packages which have fixed unit ids, so by checking the cmm dumps we can find the fully Z-encoded symbols we need, hard-code the symbols and be confident they stay the same throughout recompilation.

For non wired-in packages, however, GHC & Cabal generate unit ids with an abi hash suffix, and the hash can change when the dependencies are refactored. We can no longer hard-code those symbols. So, we need to obtain the unit id given a package name.

## Failure

What about adding `-this-unit-id` to `ghc-options`? Booting seems to work, but later compiling will fail with cryptic "bad interface" errors. It seems this flag is only usable for the wired-in packages.

## Another failure

What about extending the set of wired-in packages in GHC? Adding extra hook to override wired-in package set fails with cyclic dependency error, and hard-coding the extra packages we need will make GHC fail to compile.

Conclusion: fixed unit ids for non-wired-in packages is hard.

## Progress

In our own ghci dynamic linker, we can lookup the `ghci` package unit id and then calculate the closure symbols of asterius TH runner functions. This approach works for the TH case because:

* It happens in the host `ahc` compiler process, during TH logic, so naturally we have a valid GHC session in scope.
* The `ghci` package is guaranteed to be linked before querying the unit id.

## Challenge

What if the query needs to happen in a non-`ahc` process? Say, `ahc-ld` (or `ahc-iserv` in the future). We don't have a valid GHC session.

The most intuitive thing to do: just start a GHC session then. We only need the "global libdir" as the input to `runGhc`, which exists in `Asterius.BuildInfo`. This may work for non-`ahc` processes, but if we want to make it the universal approach of querying unit ids everywhere, another problem rises: it won't work for `ahc`. GHC has some process-global state which still prohibits multiple GHC sessions in a same process, and during TH, we will be launching GHC sessions within an already running session. So, I'm hesitant to implement this. And GHC sessions are heavy-weight and shouldn't be launched simply to calculate a symbol IMHO.

For the `ahc-ld` use case, we can consider moving linking logic back to the host `ahc` process, so there's a GHC session again. But Cabal may still call `ld` so a valid `ld` wrapper can't be taken away.

## Hack

[...]

PR #768, opened 2020-08-31
Remove redundant unit-id related logic & misc other ahc-boot fixes

TerrorJack:
* After numerous experiments, burying a black cat & drawing a pentagram, I finally managed to fix unit-id of several packages without getting cryptic bad interface errors: `unix`, `ghci` and `asterius-prelude`. Previously the unit ids would contain ABI hashes, so if we had to refer to a symbol in those packages, we couldn't embed the Z-encoded symbol name directly in the compiler/runtime source code, and a runtime query into the global package database was required. Now, the unit-ids are fixed, and we could shave off relevant logic to make things simpler & faster.
* Reduce boilerplate code in `boot.sh`


TerrorJack:
Rest in peace #332!

PR #771, opened 2020-09-08
Revamp barf logic

### LINKED ISSUE #757: Unfound symbol error message has regressed
**Describe the bug**
Previously, when `--verbose-err` is enabled for `ahc-link`, when a C function symbol is not found, attempts to call that function at runtime will result in an error message containing the function name. This is implemented with the `Barf` construct in IR and related handling in the linker.

Recently, this feature has regressed. The error message no longer contains the unfound function symbol name.

**To Reproduce**
```haskell
main :: IO ()
main = yolo

foreign import ccall unsafe "yolo" yolo :: IO ()
```

Running `ahc-link --verbose-err --run --input-hs test.hs` for the example above, the error message is something like:

```
test: JSException "RuntimeError: unreachable\n    at Main_main1_entry (<anonymous>:wasm-function[29]:0x279a)\n    at scheduleTSO (<anonymous>:wasm-function[2426]:0x8045e)\n    at scheduleTSO_wrapper (<anonymous>:wasm-function[2427]:0x8048b)\n    at Scheduler.tick (file:///workspaces/asterius/test/rts.scheduler.mjs:346:22)\n    at Immediate.<anonymous> (file:///workspaces/asterius/test/rts.scheduler.mjs:381:29)\n    at processImmediate (internal/timers.js:458:21)"
```

**Expected behavior**
The error message should contain the `yolo` string.

**Environment**
 - OS name + version: Docker dev image
 - Version of the code: latest `master` revision

**Additional context**

gkaracha:
Closes #757, following the strategy explained here: https://github.com/tweag/asterius/issues/757#issuecomment-680248390.

* Adds function `barf_push` alongside `barf` in `rts.exception.mjs`. `barf_push` takes (the ascii code of) a single character and pushes it on an internal queue. Then, function `barf_throw` reads that queue and throws the error.
* Passes `verbose_err` to both the `Binaryen` and `WasmToolkit` backends, so that we can translate `Barf`s elaborately or not, **at link-time**.
* Replaces the implementation of `Asterius.Internals.Barf.barf`: now we generate a series of `barf_push` calls, followed by a call to `barf_throw`.
* Removes `Asterius.Passes.Barf.processBarf`; no more need for `__asterius_barf_*`-style statics.
* Removes all logic relating to `__asterius_barf_*`-style statics.
* Moves all `barf`-related built-ins into `Asterius.Builtins.Barf`.


gkaracha:
@TerrorJack I started working on #757; I wonder which part of the code you meant by this in [your previous comment](https://github.com/tweag/asterius/issues/757#issuecomment-680248390):

> * Keep the old `barf` runtime function since it's used by runtime cmm files; add a new `barf_push` function

Which runtime cmm files do you mean? The way I see it, we could just replace `barf`'s implementation with that of `barf_push` and have only one barf function (and store no messages in segments, ever). Am I missing something?

TerrorJack:
@gkaracha See e.g. https://github.com/tweag/asterius/blob/master/ghc-toolkit/boot-libs/rts/StgMiscClosures.cmm, there are `barf` usages there. The Cmm string literal gets translated to a data segment anyway, so let's keep the old `barf` for now. Replacing `barf` usage with `barf_push` is hard to do at compile-time since the codegen doesn't have sufficient context to read the error message string, and as for doing it at link time...I don't think we want to further complicate the linker at this point.

gkaracha:
@TerrorJack I see. Then we'd still have to keep track of the error statics :confused: If all the old logic for the old barf must stay, then I see no merit in adding a second approach (`barf_push`). I'd close this PR and #757 and stick with what we have.

TerrorJack:
@gkaracha The existing `barf` calls don't generate data segment with `__asterius_barf` symbol names, so we can still drop the logic which deals with those symbols.

gkaracha:
@TerrorJack this should be ready for reviewing now. Two things: (a) I have left a small TODO about freezing in case there is an easy workaround, and (b) In retrospect I think that it would be a tad cleaner to have two separate functions instead of `barf_push` behaving differently depending on the input ascii. Something like replacing
```javascript
  barf_push(c) {
    if (c === 0) { // String.fromCharCode(c) === '\x00'
      throw new WebAssembly.RuntimeError(`barf_push: ${this.errorBuffer}`);
    } else {
      this.errorBuffer += String.fromCharCode(c);
    }
  }
```
with
```javascript
  barf_push(c) {
      this.errorBuffer += String.fromCharCode(c);
  }
```
and
```javascript
  barf_throw() {
    throw new WebAssembly.RuntimeError(`barf_push: ${this.errorBuffer}`);
  }
```
There is no need for NUL-terminated strings this way, and the generated barf is simply n `barf_push`es followed by one `barf_throw`, and an `Unreachable`. Let me know what you think!

TerrorJack:
@gkaracha Yes, separating `barf_push`/`barf_throw` would make things cleaner, let's do this :)

As for freezing, we can replace it with `Object.seal(this)`, so new properties can't be added after constructing the object, but existing properties can be reassigned. It's just a minor runtime mechanism to prevent accidentally using unspecified properties.

gkaracha:
> @gkaracha Yes, separating `barf_push`/`barf_throw` would make things cleaner, let's do this :)

:+1: Shouldn't take more than a minute :slightly_smiling_face: 
 
> As for freezing, we can replace it with `Object.seal(this)`, so new properties can't be added after constructing the object, but existing properties can be reassigned. It's just a minor runtime mechanism to prevent accidentally using unspecified properties.

Ah, nice, I didn't know of it. Might be a minor check but sounds nice; best to keep things as tight as possible. I'll use it :+1:

asterius/src/Asterius/Builtins.hs
```diff
@@ -163,6 +163,8 @@ rtsAsteriusModule opts =
     <> tryWakeupThreadFunction opts
     <> raiseExceptionHelperFunction opts
     <> barfFunction opts
+    <> barfPushFunction opts
```
TerrorJack, line 166: `Asterius.Builtins` has gotten quite big, so it would be nice if we can add a new module to the `Asterius.Builtins.*` namespace, and refer to that module here.

TerrorJack (changes requested):
Mostly lgtm, nice work :)

asterius/rts/rts.exception.mjs
```diff
@@ -152,4 +142,36 @@ export class ExceptionHelper {
...
+      ascii code of) a single character of the error message, followed by a
+      call to `barf_throw`.
+
+    In the backends (`Asterius.Backends.*`), when we encounter an unresolved
+    symbol `sym`, if @verbose_err@ is on, we insert a `barf` call there. So
+    if an execution path leads to the unresolved symbol, we're likely to get
+    the symbol name from the js error message.
+  */
+  barf_push(c) {
+      this.errorBuffer += String.fromCharCode(c);
```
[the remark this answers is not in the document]
gkaracha: Didn't know about `fromCodePoint`, thanks! :+1:

asterius/src/Asterius/Builtins.hs
gkaracha: Agreed. I'd name the new module `Asterius.Builtins.Barf` and I'd put in it `barf`, `barf_push`, and `barf_throw` (and the respective imports). Sounds good?

gkaracha: Just pushed a draft.

TerrorJack (approved):

PR #776, opened 2020-09-11
Implement & use safeFromIntegral

TerrorJack:


gkaracha:
@TerrorJack Perhaps relevant, I came across this package relating to safe conversions between integers (the static conversions come for free): https://hackage.haskell.org/package/int-cast-0.2.0.0/docs/Data-IntCast.html

PR #777, opened 2020-09-11
Generalize DataSegment

gkaracha:
Binaryen allows for the offset in segments to be an arbitrary expression, but until now the Asterius IR would only allow `Int32` to be used here. This PR lifts this restriction. This change should allow us to change the offset dynamically in the future (and thus unblock #750).

### DIFF
--- a/asterius/src/Asterius/Backends/Binaryen.hs
+++ b/asterius/src/Asterius/Backends/Binaryen.hs
@@ -601,7 +601,7 @@ marshalMemorySegments mbs segs = do
         =<< for
           segs
           ( \DataSegment {..} ->
-              flip runReaderT env $ marshalExpression $ ConstI32 offset
+              flip runReaderT env $ marshalExpression offset
           )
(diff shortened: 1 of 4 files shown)


gkaracha:
@TerrorJack this should unblock #750.

As an aside: I now notice that there is quite some duplication by now in `Asterius.Backends.WasmToolkit`: 4 out of 5 `make*Section` functions construct the reader environment internally (only `makeExportSection` does not require the marshaling environment in any way) . Perhaps we should create the env in `makeModule` instead to avoid this boilerplate?

TerrorJack:
Yes, let's do that.

TerrorJack (approved):
lgtm, let's wait for regular tests.

gkaracha:
> Yes, let's do that.

Do you think it's best in a separate PR or in this one?

TerrorJack:
Let's land the changes in this one.

TerrorJack (approved):

PR #778, opened 2020-09-14
Task renaming and addition of --pic flag

gkaracha:
This PR

1. Changes some unintuitive naming:
    * Renames `Task` to `LinkTask` (these tasks are handled by `ahc-link`)
    * Renames `LinkTask` to `LoadTask` (these tasks are handled by `ahc-ld`)

2. Adds a new flag, `--pic`, which instructs Asterius to generate position independent code. Note that support for position independent code is not fully implemented yet, but this change in the pipeline allows us to continue its development without users having to pay the overhead introduced by https://github.com/tweag/asterius/pull/736.


gkaracha:
@TerrorJack this should be ready for review now. Two things I am not entirely sure about:

1. Whether we should go for intrusive (but consistent) renaming (e.g. also renaming `defTask` to `defLinkTask`) or shallow.
2. Whether the bikeshedding renamings should go in a separate PR.

asterius/src/Asterius/Main.hs
```diff
@@ -264,6 +265,7 @@ ahcLink task = do
          ]
       <> ["-optl--no-gc-sections" | not (gcSections task)]
       <> ["-optl--verbose-err" | verboseErr task]
+      <> ["-optl--pic" | pic task]
```
TerrorJack, line 268: The `pic` flag is unused by `ahc-ld`, and not present in `LoadTask` right? So I don't think we need to pass `--pic` to `ahc-ld` here.

TerrorJack (changes requested):

TerrorJack:
It's fine to keep the bikeshedding renamings in this PR for now.

asterius/src/Asterius/Main.hs
gkaracha: Ah, right, that should have been part of fd0b4d545973014eb58867e870e2b186c9dcb0e8 but I missed it, nice catch!

gkaracha:
> It's fine to keep the bikeshedding renamings in this PR for now.

Agreed. What about the other part

> Whether we should go for intrusive (but consistent) renaming (e.g. also renaming defTask to defLinkTask) or shallow.

I'd say we can get away with the current renaming only, unless you have an objection.

asterius/src/Asterius/Main.hs
gkaracha: Should be fixed now.

TerrorJack:
I'm okay with the current renaming only.

TerrorJack (changes requested):
I think some changes in #750 should get in here as well, specifically: removal of "starting addresses" when making data/function offset tables, all offsets start from zero. And then, when PIC switch is off, we emit the symbols as `TAG(base + offset)`, where `base` is a hard coded value. This is important to prevent bitrotting of #750. What do you think?

TerrorJack:
Come to think of it, maybe we should just #750 into here, and leave a TODO explicitly stating that PIC mode is still WIP.

gkaracha:
> I think some changes in #750 should get in here as well, specifically: removal of "starting addresses" when making data/function offset tables, all offsets start from zero. And then, when PIC switch is off, we emit the symbols as `TAG(base + offset)`, where `base` is a hard coded value. This is important to prevent bitrotting of #750. What do you think?

Agreed! I think a few more things can be merged indeed.

> Come to think of it, maybe we should just merge #750 into here, and leave a TODO explicitly stating that PIC mode is still WIP.

This I am not entirely sure about, because #750 changes the rts too. I can start porting parts of #750 here though and see how far I can go.

TerrorJack:
> This I am not entirely sure about, because #750 changes the rts too. I can start porting parts of #750 here though and see how far I can go.

Yes, this makes sense. The `addressOf` related changes can be left there for the moment.

gkaracha:
The `--pic`-related infrastructure is integrated into #750; closing this one.

PR #790, opened 2020-09-28
Align data segments by 8 bytes instead of 16

TerrorJack:
We used to align data segments by 16 bytes. It's a simple upper-approximation since on x64 `ReadOnlyData16` requires 16-byte alignment while others require at most 8-byte. Given `ReadOnlyData16` doesn't seem to be actually used in our pipeline, this PR reduces the alignment value to 8 bytes when laying out data segments. This will result in more compact static memory data (even so when #750 lands and `--pic` is enabled).

### DIFF
--- a/asterius/src/Asterius/Passes/DataOffsetTable.hs
+++ b/asterius/src/Asterius/Passes/DataOffsetTable.hs
@@ -24,10 +24,10 @@ import Data.Tuple
 import Foreign
 import Language.Haskell.GHC.Toolkit.Constants
 
--- | Segments are 16-bytes aligned.
+-- | Segments are 8-bytes aligned.
 {-# INLINE segAlignment #-}
 segAlignment :: Int
-segAlignment = 16
+segAlignment = 8
 
 {-# INLINEABLE sizeofStatic #-}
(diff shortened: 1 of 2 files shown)


asterius/src/Asterius/Passes/DataOffsetTable.hs
```diff
@@ -35,7 +35,7 @@ makeDataOffsetTable :: AsteriusModule -> (SM.SymbolMap Word32, Word32)
 makeDataOffsetTable AsteriusModule {..} =
   swap $
     SM.mapAccum
-      ( \a ss -> (a + fromIntegral (fromIntegral (sizeofStatics ss) `roundup` 16), a)
+      ( \a ss -> (a + fromIntegral (fromIntegral (sizeofStatics ss) `roundup` 8), a)
```
gkaracha, line 38: If #750 lands first this should be equivalent to setting [`segAlignment`](https://github.com/tweag/asterius/pull/750/files#diff-a18395ea112d86e48e2e7af6b3b5eac8R27-R30) to `8`.

TerrorJack:
@gkaracha Could you update this PR since we just merged #750? Thanks :)

PR #858, opened 2021-07-04
update Github Actions

peterbecich:
https://github.com/actions/setup-haskell has been archived.
I believe https://github.com/haskell/actions/tree/main/setup is current.

-----------
Looks like there is a failure; I don't know the solution:
https://github.com/peterbecich/asterius/actions/runs/999002095


andreabedini:
The workflow has since been rewritten to not use `actions/setup-haskell` anymore. @peterbecich thanks for your contribution and apoligies for the delay.

PR #872, opened 2021-10-27
Misc simplifications

TerrorJack:
Primary goal: get rid of implicit i64/f64 castings across FFI boundaries.


TerrorJack:
Closed in favor of #876

PR #885, opened 2022-01-12
Docs: fix build, use mdbook

TerrorJack:
We used to use `sphinx`+`recommonmark` for building docs. Problem is, `recommonmark` is deprecated, and although new solutions for supporting markdown in `sphinx` exist, markdown has never been a first class citizen in `sphinx` land anyway.

This PR fixes the docs build:

* We switch to `mdbook` now. First class support for markdown, and the generated pages support automatic dark mode, which is good for my soring eyes
* The docs build is now just a plain nix build. Sans examples, which still needs to be fetched from s3 for now
* After docs build finishes, click the commit status icon, there'll be a "docs deploy" entry that links to netlify deploy URL. No need to manually find that URL in CI log


nrnrnr:
@TerrorJack This happened a little fast for my taste.  Next time you want to pull the rug out from under me, please alert first.

Also, a quick Google search showed at least four static-site generators based on pandoc.  Auto-dark mode should be possible to add...

PR #887, opened 2022-01-12
placate mdbook

nrnrnr:
mdbook sucks


purcell:
Probably the mdbook devs are very nice people, also working in the open, and when there's frustration (this is software, after all!) I might suggest it would be preferable to express it constructively. ❤️

nrnrnr:
Good point.  I'm used to private repositories, where steam can be released without an audience.

